PR #22, opened 2018-06-20
Add buildWheel task to python project

ahggns:
Maybe not totally ready to go, but curious to get input on direction


ferozco (commented):
Can we also add tests for this? Might require a change to the circle image to get python and pip around

ferozco (approved):

PR #39, opened 2018-08-28
[improvement] Gradle 4.9 and lazy task creation

dansanduleac:
## Before this PR

Was creating tasks eagerly, which makes it slower

## After this PR

Plugin creates tasks lazily, speeding up the configuration stage when these tasks are not used.

Plugin now requires gradle 4.9+


iamdanfox (changes requested):
Requiring 4.9 seems overly severe. I think we need to either find a way to give old projects old behaviour and new ones new behaviour, or just shelve this change.

PR #53, opened 2018-10-17
Cache python, typescript name and version

dansanduleac:
## Before this PR

`compileConjurePython` produces different outputs based on project name and version. However, if only these change, the task is still considered up to date.

Same for `compileConjureTypeScript`.

## After this PR

Either task will run again if you change version / name.


ferozco (commented):
https://github.com/palantir/gradle-conjure/blob/develop/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java#L38

ferozco (approved):

PR #56, opened 2018-10-25
[fix] Add guava dependency when using retrofitListenableFutures

pkoenig10:
## Before this PR
The generated retrofit API project does not include a guava dependency when using `ListenableFuture`. This requires consumers to manually add the dependency

## After this PR
The conjure plugin now automatically adds the guava dependency when `ListenableFuture` is used.


iamdanfox (changes requested):
I think this PR bumps up against a structural problem of gradle-conjure & conjure-java. At the moment conjure-java is the source of truth for parsing feature flags, so I don't really want that code to leak into gradle-conjure.

Longer-term, I think we really need to figure out a way for conjure-java to become the source of truth of dependencies.

While this is a non-default feature flag, I think it's OK to require people to do the extra manual action and take a compile dependency on guava.

pkoenig10:
@iamdanfox now that https://github.com/palantir/conjure-java/pull/641 has merged, would you be open to re-opening the PR and unconditionally adding this dependency? Otherwise clients are going to fail to compile when conjure-java is upgraded.

iamdanfox:
Yep, now that it's a default feature flag I think this is what we want.

PR #59, opened 2018-10-29
[fix] Refactor to service dependencies

ferozco:
## Before this PR
Exposed a `productDependencies` extension which allow you to define a set of `productDependency` 
ex.

```
productDependencies {
  productDependency {
    productGroup = "com.palantir.conjure"
    productName = "conjure"
    minimumVersion = "1.2.0"
    recommendedVersion = "1.2.0"
    maximumVersion = "2.x.x"
  }
}
```

## After this PR
Renamed extensions to more closely align with the conjure model of objects and services.
ex.
```
serviceDependencies {
  serviceDependency {
    productGroup = "com.palantir.conjure"
    productName = "conjure"
    minimumVersion = "1.2.0"
    recommendedVersion = "1.2.0"
    maximumVersion = "2.x.x"
  }
}
```

Open Question: 
Should we follow up with `conjure-typescript` to push the refactor further?

### DIFF
--- a/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureProductDependenciesExtension.java
+++ b/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureProductDependenciesExtension.java
@@ -23,18 +23,18 @@ import java.util.Set;
 
 public class ConjureProductDependenciesExtension {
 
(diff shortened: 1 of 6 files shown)


iamdanfox (approved):
Looks fine to me, and ensures we don't clash with sls-packaging itself!

PR #65, opened 2018-11-06
[wip] [improvement] Provide default versions for tool dependencies

dansanduleac:
## Before this PR

It's hard to set up a gradle project from scratch.
According to the [getting started guide](https://github.com/palantir/conjure/blob/develop/docs/getting_started.md), you have to specify versions for each generator manually, either by 
* replacing the versionless dependencies that are auto-added by gradle-conjure with the same dependency but specifying an actual version
* using `nebula.dependency-recommender` to provide the versions

## After this PR

We are now using [configuration defaults](https://docs.gradle.org/current/userguide/customizing_dependency_resolution_behavior.html#sec:configuration_defaults) which allow you to provide default versions for these tools, whilst allowing you to override them with your own dependencies, ~as well as forcing their versions with nebula~.

TODO:

- [ ] figure out how to allow nebula to force dependencies coming from `defaultDependencies`


iamdanfox:
Closing out for tidiness.

PR #71, opened 2018-11-13
Set up publishing to gradle

dansanduleac:


iamdanfox (approved):
LGTM - let's just wait until Gradle approves our request and this actually publishes!

derenrich:
bump on this? it's blocking a plugin I'm writing

dansanduleac:
@derenrich I couldn't figure out a way to publish multiple jars through gradle's plugin (we need to publish the api jar to the same repo somehow, since it's a dependency of the plugin jar).
Either we get rid of the API jar (which is currently being used internally), or this is impossible.

iamdanfox:
Closing out stale PR

Issue #97, opened 2019-01-07
support publish conda task


derenrich:
## What happened?

There's a publish npm task for typescript but no corresponding task for python.

## What did you want to happen?

A way to publish python code to a conda or pypi repository

iamdanfox:
Closing out as I don't think this plugin is the best home for that functionality.  Would recommend putting this in our internal publish plugin, or perhaps the conda one.

PR #98, opened 2019-01-07
[improvement] add publishWheel task and tests

derenrich:
## Before this PR
I experienced the issue as described in https://github.com/palantir/gradle-conjure/issues/97
## After this PR
Now you can run `./gradlew publishWheel` to publish the wheel built via `./gradlew buildWheel` using `twine`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,8 +17,9 @@ jobs:
       - restore_cache: { key: 'gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
       - restore_cache: { key: 'gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
       - run: sudo apt-get update
-      - run: sudo apt-get install twine python-setuptools python-pip
-      - run: ./gradlew --no-daemon --parallel --stacktrace  --continue --max-workers=8 --profile build --debug
+      - run: sudo apt-get install python-setuptools python-pip
+      - run: pip install twine
+      - run: ./gradlew --no-daemon --parallel --stacktrace  --continue --max-workers=8 --profile build
       - save_cache:
           key: 'gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}'
           paths: [ ~/.gradle/wrapper ]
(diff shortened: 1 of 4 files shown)


iamdanfox:
The PR itself is very tidy, but I'm a bit uncomfortable about scope creep of the gradle-conjure plugin.  

The original intention of this repo was to stay as simple as possible and just delegate to the respective generators, ideally containing no special-casing code (assuming RFC002 only) so that we keep all logic in generators and another build tool could be implemented easily.

Clearly there's already been a bit of scope creep because we do have a `publishTypeScript` task which relies on the existence of `npm` to call `npm run build` and `npm publish`.  I think this is just about acceptable because npm is very likely to be pre-installed and should work out of the box.  I'm less comfortable about relying on [`twine`](https://pypi.org/project/twine/) as it might not be installed out of the box.

I see a few options:

- just merge this PR as is, reason that gradle is usually responsible for build _and publishing_, and typescript already set a precedent.  Conjure maintainers accept the risk of a ballooning language-specific support load as more language authors add publishing logic to gradle-conjure.
- adjust the PR to just use `python setup.py upload` and just rely on [conjure-python](https://github.com/palantir/conjure-python/) generating something sensible
- reject the extra task and take a hard line of no more language specific stuff in gradle-conjure.  Provide documentation with a copy&pasteable gradle snippet for people to do publishing, e.g.:
```gradle 
tasks.register("publishWheel", Exec) {
  dependsOn buildWheelPython
  // ...
}
```

.circleci/config.yml
```diff
@@ -16,6 +16,9 @@ jobs:
       - checkout
       - restore_cache: { key: 'gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}' }
       - restore_cache: { key: 'gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}' }
+      - run: sudo apt-get update
+      - run: sudo apt-get install python-setuptools python-pip
+      - run: sudo pip install twine
```
iamdanfox, line 21: These lines diverge from the .circleci/config.yml template, so they'll be overwritten by excavator.  We'd need to brainstorm a slightly different way of doing python integration tests, or possibly upstream something into the template.

derenrich:
So using `setup.py upload` was the original path I followed in this PR but it emitted warnings because it is deprecated. I would be fine going back to it though despite the warnings but we would still need some packages installed (e.g. setuptools) which kinda defeats the point.

Honestly what we really need is conda publishing and I was going to add that next but that requires even an additional dependency on conda.

I would personally prefer swapping out the docker image itself to add python/twine but that seemed more involved/controversial.

.circleci/config.yml
derenrich: i'm unfamiliar with this templating/excavator system. Where does the template live and how is it controlled?

iamdanfox:
One alternative would be to extend the existing gradle-conda plugin we have internally to know about gradle-conjure??  How important is it that this functionality is OSS?

derenrich:
so you're imagining users of the current plugin hosted here would switch to just the python specific one if they additionally wanted python publishing? or do you imagine having both of them?

the former seems weird to me. really i just want an easy way to convince everyone to publish python bindings for their libraries with minimal code changes/ugliness.

iamdanfox:
> i just want an easy way to convince everyone to publish python bindings for their libraries with minimal code changes/ugliness.

if it's acceptable to just get internal users publishing their python bindings then I think doing something like:

```diff
 apply plugin: 'com.palantir.conjure'

 projects(':foo-api-python) {
     apply plugin: 'com.palantir.conda.base'
    ...
+    apply plugin: 'com.palantir.publish-conda' // this would be a new plugin that we can publish from our existing internal repo?
 }
```

derenrich:
i'd be ok with that though it'd probably be called `publish-conjure-conda`. I'm indifferent to open sourcing it. Probably be easier not to.

PR #116, opened 2019-03-05
[improvement] Stop using lambda in doFirst to improve build caching

lycarter:
## Before this PR

Copying conjure sources was not cacheable because gradle does not (currently) support Java lambdas: https://github.com/gradle/gradle/issues/5510

bluf of the issue: Java lambdas do not reproducibly generate the same ID internally, so gradle is unable to determine if the lambda has changed. The easy fix is to just use an anonymous function until gradle fixes this upstream.

## After this PR

Copying conjure sources is now cacheable

relevant to https://github.com/gradle/gradle/issues/5510

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.Set;
 import java.util.function.Supplier;
 import javax.inject.Inject;
+import org.gradle.api.Action;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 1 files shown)


carterkozak:
@lycarter Any chance you can write a test that fails without this, and passes with the fix? That way we can be confident we don't reintroduce this break. We tests for similar problems, see `compileConjure does not run tasks if up to date` in `ConjurePluginTest.groovy`

lycarter:
@cakofony this task is already in that test: https://github.com/palantir/gradle-conjure/blob/develop/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy#L241

The test passes on Gradle 4.10.2 (what this repo is on), and I believe will fail on Gradle >= 5.0, per some of the comments on that linked gradle issue. I tried testing that out to confirm, but a large pile of other tests fail when I tried upgrading to 5.1 :/

I can dig into upgrading to upgrading to >=5.0, but it'll probably take a little while

lycarter:
Dug in a bit further on the issues with upgrading to Gradle 5.0: they all seem to be because of methods that were deprecated in Gradle 5.0, and there's a check to ensure no deprecated methods are used. Most of these methods are pulled in from `gradle-baseline`, which is on 5.2 but doesn't have the deprecation check :(

lycarter:
lame, a Gradle 5 upgrade for this repo (while deprecation checks are enabled) is blocked on https://github.com/palantir/gradle-baseline/pull/493 :(

jdm2212:
@cakofony this PR can't be tested because it's fixing an issue with a future version of Gradle (and, once you upgrade to that future version, this will be tested by existing tests). In the interim, the caching problem is making `gw checkstyleMain` take ~40s on our internal project (your old project, incidentally :) we can share build scans on internal comms), so it'd be great if we could merge this and get a new release so that we can cut down our build times.

carterkozak:
I defer to the other service infnra folks who have stronger gradle-fu than myself :-) The change looks reasonable to me though!

iamdanfox (approved):
:+1: thanks @lycarter, I did a similar thing recently (https://github.com/palantir/gradle-conjure/pull/108) but clearly missed this case!

PR #120, opened 2019-03-14
[fix] Relax checkConjureJavaVersions to not include major version check

rubenfiszel:
This will allow repo to upgrade WC through the excavator thanks to gradle-consistent-version and then upgrade conjure-java.


iamdanfox:
:+1:

PR #141, opened 2019-04-19
[improvement] All tasks are now compatible with the Gradle Build Cache

iamdanfox:
## Before this PR

Using gradle enterprise, I noticed that a decent chunk of time on every CI build is spent re-running our conjure generators.  These are particularly slow because we require the abstraction of a tgz, so we pay the price of a JVM startup.

## After this PR
==COMMIT_MSG==
./gradlew compileConjure is now compatible with the build cache
==COMMIT_MSG==

This should mean that after a `git clean -xdf` (or switched local branches), a second run of `./gradlew compileConjure` should be _much_ faster, with lots of tasks showing `FROM-CACHE`. Retriggered builds on CI should also be faster.

## Possible downsides?
If someone implements a generator that is not deterministic, this might give them confusing behaviour.  Also, if don't declare inputs correctly then this functionality might annoy users who change source files but don't see their tasks re-run. Thankfully, using `--rerun-tasks` is an easy workaround here

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -69,7 +69,7 @@ public class CompileIrTask extends DefaultTask {
         getProject().exec(execSpec -> {
             ImmutableList.Builder<String> commandArgsBuilder = ImmutableList.builder();
             commandArgsBuilder.add(
-                    executableDir.get().toPath().resolve("bin").resolve("conjure").toFile().getAbsolutePath(),
+                    new File(executableDir.get(), "bin/conjure").getAbsolutePath(),
                     "compile",
                     inputDirectory.get().getAbsolutePath(),
                     outputFile.getAbsolutePath());
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -21,16 +21,19 @@ import java.io.File;
 import java.util.List;
 import java.util.function.Supplier;
 import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.InputDirectory;
-import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.OutputFile;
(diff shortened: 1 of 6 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -46,25 +49,27 @@ public final File getOutputFile() {
...
+        return executableDir.get();
     }
 
     @TaskAction
     public final void generate() {
         getProject().exec(execSpec -> {
             ImmutableList.Builder<String> commandArgsBuilder = ImmutableList.builder();
             commandArgsBuilder.add(
-                    executablePath.get().getAbsolutePath(),
+                    executableDir.get().toPath().resolve("bin").resolve("conjure").toFile().getAbsolutePath(),
```
carterkozak: Odd to switch from File->Path->File.`toAbsolutePath().toString()` works, or `new File(executableDir.get(), "bin/conjure").getAbsolutePath()` (even on windows with modern javas!)

carterkozak (commented):
Is there any way we can test this similarly to our up-to-date tests?

iamdanfox:
I've validated this works using Gradle Enterprise and just running things twice, but agreed it's probably worth covering these in some decent tests just so they don't regress!

As usual Gradle has pretty decent testing infra available: https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-build-cache (although I think we'll need to use their clean directory between runs snippet).

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java
```diff
@@ -60,6 +64,7 @@ final void setExecutablePath(Supplier<File> executablePath) {
     }
 
     @InputFile
+    @PathSensitive(PathSensitivity.NONE)
```
ferozco, line 67: Why did we make this path insensitive? Doesn't the path change when the executable changes?

iamdanfox: We'd want to use one of the others if the path name could be the _only_ thing that changes.  I don't think that's every the case though, as the contents of the file is good enough to determine whether we should re-run or not

ferozco:
👍

PR #144, opened 2019-05-22
[improvement] Allow -jaxrs suffix

iamdanfox:
## Before this PR

As part of the migration to undertow, people are getting slightly confused about whether they can use undertow on the client-side and whether they can fully delete the -jersey project.

-jersey is actually not quite right because it only has jaxrs annotations (e.g. `@Path`, `@GET` etc), which is why it _does_ work on both the client-side and the server-side!

## After this PR
==COMMIT_MSG==
Naming a project :foo-jaxrs will now get the same codegen as :foo-jersey
==COMMIT_MSG==

## Possible downsides?

It's a bit hacktastic, haven't renamed everything under the hood yet.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -25,6 +25,7 @@ import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Set;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 import javax.inject.Inject;
 import org.gradle.api.Action;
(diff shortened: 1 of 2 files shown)


iamdanfox:
WOMP https://status.circleci.com/

Screenshot 2019-05-22 at 17 49 27

carterkozak:
@iamdanfox Do you think we should provide a client-specific flavor of the jaxrs generator? e.g. https://github.com/palantir/conjure-java/pull/309 (requires some heavy rebasing)

iamdanfox:
Ah good point, I don't want to take over the -jaxrs name with the old one when we have the shiny new one: https://github.com/palantir/conjure-java/pull/309

PR #146, opened 2019-05-31
Support first-class Java projects in ConjureLocalJavaPlugin

uschi2000:


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
```diff
@@ -127,10 +129,39 @@ private void setupGenericConjureProjects(
                                         task.setSource(conjureIrConfiguration);
                                         task.setExecutablePath(extractConjureGeneratorTask::getExecutable);
                                         task.setOptions(() -> conjureExtension.getGenericOptions(subprojectName));
-                                        task.setOutputDirectory(subproject.file(subprojectName));
+
+                                        final String outputDir;
+                                        if (conjureExtension.getGenericOptions(subprojectName).has("outputDir")) {
```
uschi2000: @ferozco this does not get evaluated lazily, i.e., I don't have access to options at this point in the code. any idea why? your recommendation for working around this?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
```diff
@@ -17,15 +17,26 @@
...
 import java.io.File;
+import java.util.function.Function;
 import org.gradle.api.tasks.CacheableTask;
 
 @CacheableTask
 public class ConjureLocalGenerateTask extends ConjureGeneratorTask {
 
+
+    // By default, strip extension and version
+    private Function<File, File> outputDirFor = file ->
```
ferozco: We may just want to have a separate task for java local codegen

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
ferozco: I would recommend to support Java as a first class language and wire it up seperately from the generic generators. I think this is desirable since you'll also will have to configure source sets/idea modules/dependencies for the java project

uschi2000:
@ferozco ready for you

ferozco (commented):
It seems like the new plugin is in an odd spot, partially copying the the pattern of generic generators from `conjure-local` while adding some special java-specific semantics. I feel like this stems from a tension of wanting to support Java in a first class way while being unwilling to make the simplifying assumption that there is a single generator (ex. conjure-java) that produces Java code.

I see two different approaches to solving the problem of Java local codegen:

- Add explicit support for `conjure-java` to `com.palantir.conjure-local` in the same way that `conjure-typescript` and `conjure-python`. We can borrow/re-use the dependency and project wiring from `com.palantir.conjure`.

- Commit to a completely separate `conjure-java-local` plugin which only supports Java generators. This plugin should have an entirely separate extension and could generate code within the target project instead of a child-project.


I suggest we follow the first approach. I lean towards this approach as it is simpler from an implementation perspective (minimal code change and no additional plugins) and will bring `com.palantir.conjure-local` in line with `com.palantir.conjure`. It also has the added benefit of continuing to support `conjure-java` as the source of truth for Java code gen.

I think this approach also addresses the need to support experimental generators since with a small gradle snippet you can replace `conjure-java` with the generator of your choice.

```gradle
configurations.conjureJava {
    resolutionStrategy.dependencySubstitution {
        it.substitute it.module('com.palantir.conjure.java:conjure-java') with it.module('com.palantir.generator:my-awesome-generator:1.0.0')
    }
}
```

uschi2000:
Well, the assumption `that there is a single generator (ex. conjure-java) that produces Java code` is not true, so I'm unwilling to make it. The second approach is what this PR does.

iamdanfox:
Closing out in favour of https://github.com/palantir/gradle-conjure/pull/147 and https://github.com/palantir/gradle-conjure/pull/148

PR #147, opened 2019-06-03
[improvement] support conjure-java in local plugin

ferozco:
## Before this PR
We did not support Java as a first class language within `com.palantir.conjure-local`. Users would be surprised to find that the `java` section of the `conjure` extension was ignored and the only was to generate java code was to use the generic generator config, ex:
```gradle
conjure {
    options "java", {
      ...
    }
}
```

## After this PR
==COMMIT_MSG==
Add basic support for conjure-java in `com.palantir.conjure-local`
==COMMIT_MSG==

Since `conjure-java` requires a flag to be passed to specify what category of code to generate, it does not work out of the box.  Instead, users will have to do the following:
```gradle
conjure {
    java {
        // One of the following
        objects = true
        jersey = true
        retrofit = true
        undertow = true
    }
}
```

We can fix this by updating `conjure-java` to support multiple flags being passed at once. 

## Possible downsides?
Local code-gen is fully supported and will have rough edges that users can run into. In particular, I'm concerned about the danger of duplicate classes if a library uses this plugin and loss of automatic product dependencies.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
@@ -31,6 +31,8 @@ import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.file.FileCollection;
+import org.gradle.api.plugins.BasePlugin;
+import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.util.GUtil;
 
 public final class ConjureLocalPlugin implements Plugin<Project> {
@@ -38,14 +40,17 @@ public final class ConjureLocalPlugin implements Plugin<Project> {
     /** Configuration where custom generators should be added as dependencies. */
     private static final String CONJURE_GENERATORS_CONFIGURATION_NAME = "conjureGenerators";
 
+    private static final String JAVA_PROJECT_NAME = "java";
     private static final String PYTHON_PROJECT_NAME = "python";
     private static final String TYPESCRIPT_PROJECT_NAME = "typescript";
     private static final ImmutableSet<String> FIRST_CLASS_GENERATOR_PROJECT_NAMES = ImmutableSet.of(
-            PYTHON_PROJECT_NAME, TYPESCRIPT_PROJECT_NAME);
+            JAVA_PROJECT_NAME, PYTHON_PROJECT_NAME, TYPESCRIPT_PROJECT_NAME);
     private static final String CONJURE_GENERATOR_DEP_PREFIX = "conjure-";
 
     @Override
     public void apply(Project project) {
+        project.getPlugins().apply(BasePlugin.class);
+
         Configuration conjureIrConfiguration = project.getConfigurations().maybeCreate(CONJURE_CONFIGURATION);
         Configuration conjureGeneratorsConfiguration = project.getConfigurations().maybeCreate(
                 CONJURE_GENERATORS_CONFIGURATION_NAME);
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
```diff
@@ -66,6 +73,51 @@ public void apply(Project project) {
...
+        if (subproj == null) {
+            return;
+        }
+
+        Configuration conjureJavaConfig = project.getConfigurations().maybeCreate(ConjurePlugin.CONJURE_JAVA);
+        File conjureJavaDir = new File(project.getBuildDir(), ConjurePlugin.CONJURE_JAVA);
+        project.getDependencies().add(ConjurePlugin.CONJURE_JAVA, ConjurePlugin.CONJURE_JAVA_BINARY);
+        ExtractExecutableTask extractJavaTask = ExtractExecutableTask.createExtractTask(
+                project, "extractConjureJava", conjureJavaConfig, conjureJavaDir, "conjure-java");
+
```
uschi2000, line 93: nit: extra \n

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureLocalPluginTest.groovy
```diff
@@ -57,6 +58,25 @@ class ConjureLocalPluginTest extends IntegrationSpec {
...
+        }
+        """.stripIndent()
+
+        when:
+        ExecutionResult result = runTasksSuccessfully("generateConjure")
+
+        then:
+        result.wasExecuted(":generateJava")
+
+        fileExists('java/src/generated/java/com/palantir/conjure/spec/ConjureDefinition.java')
```
uschi2000: can also check that iml file has right contents, e.g.
```
file('java/java.iml').text.contains('src/generated/java')
```

uschi2000 (approved):

iamdanfox:
Without package relocating, I think arbitrary conjure-java local codegen is dangerous. (Anyone that locally codegens something and publishes it as part of a library will mean that all consumers then need to make sure they don't have any duplicates of that class anywhere on their classpath).

Given that we're just trying to unblock the dialogue codegen, can we just specifically whitelist dialogue for now? Either if it's just the dialogue jar, or if it gets mainlined:

```gradle
conjure {
    java {
        dialogue = true
    }
}
```

Does dialogue specifically need local codegen of objects as well? Or could we lock this down until the package relocation is implemented?

ferozco:
Agree with @iamdanfox about the danger of arbitrary codegen. I wasn't sure what was required for dialogue so I left it up to the user to configure what code was generated. Happy to restrict further if you'd like. @uschi2000 what is required for dialogue code gen? Could we configure conjure-java(-dialogue) to accept a single flag (`--dialog`) to generate all the required code?

markelliot:
> all consumers then need to make sure they don't have any duplicates of that class anywhere on their classpath

Don't we include a plugin in gradle-baseline to prevent exactly this kind of problem?

iamdanfox:
We did whip together `com.palantir.baseline-class-uniqueness` just for fun a while ago, but haven't rolled it out yet (loads of services just disable it)

carterkozak:
Thoughts on a uniqueness-exceptions.lock we can generate once to allow rollout and prevent _new_ regressions? This way way users must at the very least acknowledge when new duplicate classes are added and accept the risk.

uschi2000:
@ferozco we have two use cases with two different needs: @ellisjoe wants to generate objects and dialogue classes, I need to generate just dialogue classes.

iamdanfox:
Just chatted to @ellisjoe and his codegen isn't deeply coupled to the objects either, so I think we could just use the `-objects` jars published by an upstream service for now.

Have filed https://github.com/palantir/gradle-baseline/issues/648 to track Carter's proposal for how to get baseline-class-uniqueness rolled out more widely.

uschi2000:
ok, let's do it

ferozco:
@iamdanfox are we ok to move forward with this?

iamdanfox:
I think for maximum safety I would kinda like to blacklist the objects, jersey, retrofit, undertow options, so that this can really only be used to produce `dialogue` stuff initially. At least until we've got a compelling class uniqueness or package relocation story sorted for objects.

But yeah in general :+1:

PR #148, opened 2019-06-05
[improvement] support generic generator for API producers

### LINKED ISSUE #90: Support Conjure generators on local project
Right now if I have a project called `my-service-api`, I can create a new project called `my-service-local-api`, add `conjure-local`, pull down my published Conjure IR from `my-service-api`, and use that to generate, say, a Postman collection.

Instead, I'd like to add some plug-in(s) to my existing `my-service-api` project that give(s) me a `gradle generateConjurePostman` task that I can run right from my project to generate a Postman collection from the current local Conjure files.

cc @ferozco

### LINKED ISSUE #100: com.palantir.conjure doesn't support generic generators
It appears that only com.palantir.conjure-local supports generic generators, but that doesn't support the naming conventions that com.palantir.conjure does. It seems like com.palantir.conjure itself should support custom generators as well.

ferozco:
Closes #90 
Closes #100 

## Before this PR
API producers could only generate code using the first class generators (Java, Python, TypeScript)

## After this PR
==COMMIT_MSG==
Add support to `com.palantir.conjure` for generic Conjure generators 
==COMMIT_MSG==

## Possible downsides?
Additional code path to maintain

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -18,6 +18,7 @@ package com.palantir.gradle.conjure;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
 import com.palantir.gradle.conjure.api.ConjureExtension;
 import com.palantir.gradle.conjure.api.ConjureProductDependenciesExtension;
 import com.palantir.gradle.conjure.api.GeneratorOptions;
@@ -499,6 +500,11 @@ public final class ConjurePlugin implements Plugin<Project> {
             Task compileConjure,
             Task compileIrTask,
             Configuration conjureGeneratorsConfiguration) {
+        Map<String, Project> genericSubProjects = Maps.filterKeys(
+                project.getChildProjects(),
+                key -> !FIRST_CLASS_GENERATOR_PROJECT_NAMES.contains(
+                        extractSubprojectLanguage(project.getName(), key)));
+
         // Validating that each subproject has a corresponding generator.
         // We do this in afterEvaluate to ensure the configuration is populated.
         project.afterEvaluate(p -> {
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -536,6 +536,35 @@ class ConjurePluginTest extends IntegrationSpec {
...
     }
 
+    def 'supports generic generators'() {
+        addSubproject(':api:api-postman')
+        file('versions.props') << """
+        com.palantir.conjure.postman:conjure-postman = 0.1.0
+        """.stripIndent()
+        file('api/build.gradle') << """
+        dependencies {
+            conjureGenerators 'com.palantir.conjure.postman:conjure-postman'
```
iamdanfox, line 546: could we make this `conjureGenerator` singular, instead of plural?

iamdanfox:
I wonder if it's time to revisit the thing where the subproject `:foo-api:foo-api-postman` is entirely configured from the parent `:foo-api` project?

Especially when we're trying to wire up generators <-> subprojects, it seems like it might be less weird from an implementation perspective if the actual `foo-api-postman` project is in control of it's generator??

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -471,6 +493,77 @@ private static void setupConjurePythonProject(
...
+                    .stream()
+                    .collect(Collectors.toMap(dependency -> {
+                        Preconditions.checkState(dependency.getName().startsWith(CONJURE_GENERATOR_DEP_PREFIX),
+                                "Generators should start with '%s' according to conjure RFC 002, "
+                                        + "but found name: '%s' (%s)",
+                                CONJURE_GENERATOR_DEP_PREFIX, dependency.getName(), dependency);
+                        return dependency.getName().substring(CONJURE_GENERATOR_DEP_PREFIX.length());
+                    }, Function.identity()));
+
+            project.getChildProjects().entrySet().stream()
```
iamdanfox: rather than doing .entrySet().stream().forEach() I think you can use the native forEach method on a Map, which gives you both the key and the value as two separate args

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -471,6 +493,77 @@ private static void setupConjurePythonProject(
...
+                                && !generators.containsKey(conjureLanguage)) {
+                            throw new RuntimeException(String.format("Discovered subproject %s without corresponding "
+                                            + "generator dependency with name '%s'",
+                                    subproject.getPath(), ConjurePlugin.CONJURE_GENERATOR_DEP_PREFIX + subprojectName));
+                        }
+                    });
+        });
+
+        project.getChildProjects().entrySet().stream()
+                .filter(e -> FIRST_CLASS_GENERATOR_PROJECT_NAMES.stream()
```
iamdanfox: This whole afterEvaluate block is just working on these generic subprojects, not the first-class ones... Maybe just hoist a variable like genericSubprojects to do this once? e.g.

```
Maps.filterKeys(project.getChildProjects(), key -> !FIRST_CLASS_GENERATOR_PROJECT_NAMES.contains(key));
```

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -471,6 +493,77 @@ private static void setupConjurePythonProject(
...
+                    Project subproject = e.getValue();
+                    String conjureLanguage = extractSubprojectLanguage(project.getName(), subprojectName);
+
+                    // We create a lazy filtered FileCollection to avoid using afterEvaluate.
+                    FileCollection matchingGeneratorDeps = conjureGeneratorsConfiguration.fileCollection(
+                            dep -> dep.getName().equals(CONJURE_GENERATOR_DEP_PREFIX + conjureLanguage));
+
+                    ExtractExecutableTask extractConjureGeneratorTask = ExtractExecutableTask.createExtractTask(
+                            project,
+                            GUtil.toLowerCamelCase("extractConjure " + conjureLanguage),
```
iamdanfox: woah is that space intentional?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -471,6 +493,77 @@ private static void setupConjurePythonProject(
...
+                            matchingGeneratorDeps,
+                            new File(subproject.getBuildDir(), "generator"),
+                            String.format("conjure-%s", conjureLanguage));
+
+                    ConjureGeneratorTask conjureLocalGenerateTask = project.getTasks()
+                            .create(GUtil.toLowerCamelCase("compile conjure " + conjureLanguage),
+                                    ConjureGeneratorTask.class,
+                                    task -> {
+                                        task.setDescription(String.format(
+                                                "Generates %s files from remote Conjure definitions.",
```
iamdanfox: no longer `remote` right?

iamdanfox (approved):
Looks solid, thanks @ferozco

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
ferozco: Agree we should have done that originally. Unfortunately, It would be a break to rename the configuration in `com.palantir.conjure-local` and I think I'd prefer for the configurations to consistently named

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
ferozco: yes, its to specify the word boundary so that the resulting string is capitalized correctly

ferozco:
@iamdanfox totally agree it is a little weird from the implementation perspective. I would only want to revisit if we could make the change across the board (including first class generators), since at least for now it is consistently weird.

Issue #33, opened 2018-07-30
windows compat of compileIrTask


clockfort:
I think conjure 4.0 / https://github.com/palantir/gradle-conjure/commit/41bf5fcdb8de4e7c05b46fcf59aa3d14a548a46c#diff-1036bc6316fc17fcb315276f5008e379 might have broken Windows compat for the compileIR task, from both gradle usage in cygwin and gradle usage in native shell.

no big deal and not a blocker for me since it still works fine in my Windows Subsystem for Linux shell, but thought you might want to know about the compat break.

```
Execution failed for task ':myProject-api:compileIr'.
> A problem occurred starting process 'command 'C:\Users\clockfort\code\myProject\myProject-api\build\conjureCompiler\bin\conjure''

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':myProject-api:compileIr'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:110)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipCachedTaskExecuter.execute(SkipCachedTaskExecuter.java:105)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
        at org.gradle.api.internal.tasks.execution.ResolveBuildCacheKeyExecuter.execute(ResolveBuildCacheKeyExecuter.java:66)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:101)
[...]

PR #164, opened 2019-07-05
[improvement] Support conjure generation on windows

CRogers:
## Before this PR
Currently, conjure fails to generate on Windows as gradle-conjure always tries to execute a bash script to run IR compilation and generators. This is preventing me from using conjure in https://github.com/palantir/docker-compose-rule/pull/292 as we support develop from external developers who use windows.

## After this PR
==COMMIT_MSG==
Conjure IR generation and conjure generators will now be called by `${GENERATOR_NAME}.bat` on windows rather than simply `${GENERATOR_NAME}`. This allows windows support.
==COMMIT_MSG==

Conjure IR generation and java/python conjure generation at least already produce the required `.bat` files automatically.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/OsUtils.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/OsUtils.java
@@ -17,13 +17,13 @@
 package com.palantir.gradle.conjure;
 
 import java.io.File;
-import org.apache.commons.lang3.SystemUtils;
+import org.apache.tools.ant.taskdefs.condition.Os;
 
 final class OsUtils {
     private OsUtils() {}
 
     static String appendDotBatIfWindows(String executable) {
-        return executable + (SystemUtils.IS_OS_WINDOWS ? ".bat" : "");
+        return executable + (Os.isFamily(Os.FAMILY_WINDOWS) ? ".bat" : "");
     }
 
     static File appendDotBatIfWindows(File executable) {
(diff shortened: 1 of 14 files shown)


CRogers:
Actually fixes a regression (see #33)

gradle-conjure/src/main/java/com/palantir/gradle/conjure/OsUtils.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.conjure;
+
+import java.io.File;
+import org.apache.commons.lang3.SystemUtils;
```
iamdanfox: Isn't there a gradle-native way of checking this without requiring a new apache dependency?

iamdanfox (commented):
Looks good to me - should also mention this in the official conjure docs: https://palantir.github.io/conjure/#/docs/rfc/002-contract-for-conjure-generators.

Are we gonna say generators are not required to support windows, but the existence of this file implies they do?

iamdanfox:
@CRogers wanna lift the appveyor stuff from https://github.com/palantir/docker-compose-rule/blob/master/appveyor.yml to make sure this is actually sufficient to declare compatibility with windows?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/OsUtils.java
CRogers: Looks like there's an internal gradle API to find this, but it has changed at least twice. There's a gradle-bundled ant thing I use instead with no extra deps.

CRogers:
I would imagine it as saying that you don't need to support windows, but it's now possible.

I'll try to set up appveyor - bet it need some kind of setup through a UI though.

CRogers:
Fixes #33

dansanduleac:
👍

PR #183, opened 2019-08-21
use GCV

ferozco:
## Before this PR
We did not use GCV

## After this PR
==COMMIT_MSG==
Use GCV to manage dependencies
==COMMIT_MSG==

## Possible downsides?
N/A


iamdanfox (approved):
Could you also make sure the `publish-jar.gradle` has all the right resolutionResult stuff to make the POM look sensible? I know we forgot this when we did iceberg!

PR #201, opened 2019-09-04
Support conjure-rust

sfackler:
## After this PR
==COMMIT_MSG==
Generation of Rust code from Conjure definitions is now supported.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -58,12 +58,13 @@ public final class ConjurePlugin implements Plugin<Project> {
     public static final String CONJURE_IR = "compileIr";
 
     private static final ImmutableSet<String> FIRST_CLASS_GENERATOR_PROJECT_NAMES = ImmutableSet.of(
-            "objects", "jersey", "retrofit", "undertow", "typescript", "python");
+            "objects", "jersey", "retrofit", "undertow", "typescript", "python", "rust");
 
     // configuration names
     static final String CONJURE_COMPILER = "conjureCompiler";
     static final String CONJURE_TYPESCRIPT = "conjureTypeScript";
     static final String CONJURE_PYTHON = "conjurePython";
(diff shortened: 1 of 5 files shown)


CRogers:
Hmmm, looks like the windows build is failing. All of the rest conjure ecosystem currently supports windows - does conjure-rust do so?

sfackler:
Ah yeah we aren’t currently building a Windows binary. I’ll see about setting that up.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -487,6 +494,38 @@ private static void setupConjurePythonProject(
...
+        if (project.findProject(rustProjectName) != null) {
+            Configuration conjureRustConfig = project.getConfigurations().maybeCreate(CONJURE_RUST);
+
+            project.project(rustProjectName, subproj -> {
+                applyDependencyForIdeTasks(subproj, compileConjure);
+                File conjureRustDir = new File(project.getBuildDir(), CONJURE_RUST);
+                project.getDependencies().add(CONJURE_RUST, CONJURE_RUST_BINARY);
+                ExtractExecutableTask extractConjureRustTask = ExtractExecutableTask.createExtractTask(
+                        project, "extractConjureRust", conjureRustConfig, conjureRustDir, "conjure-rust");
+                project.getTasks().create("compileConjureRust",
```
ferozco, line 509: Out of curiosity, how are users going to consume the generated crate? What does publishing look like?

sfackler: The current support is designed for rust-based services, where your main crate would just depend on the generated Conjure crate via the local filesystem path.

Publishing support is a follow-up since it'll require a bit more work (we'll need to download a Cargo binary appropriate for the host, for example).

sfackler:
Windows build is green.

sfackler:
Turns out that the generic generator support is sufficient for everything we're doing here, so I'm going to close.

PR #202, opened 2019-09-05
Eliminate some warnings (because 'compile' is deprecated)

iamdanfox:
## Before this PR

Users of gradle-conjure may see some lame deprecation warnings:

```
> Configure project :log-receiver-api
The compile configuration has been deprecated for dependency declaration. This will fail with an error in Gradle 7.0. Please use the implementation or api configuration instead.
        at build_3eb5450vc53poxr4t2qpcbw9e.run(/Volumes/git/log-receiver/log-receiver-api/build.gradle:1)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

## After this PR
==COMMIT_MSG==
gradle-conjure produces no longer sets up deprecated 'compile' dependencies, instead preferring to declare 'api' dependencies (introduced with the java-library plugin).
==COMMIT_MSG==

## Possible downsides?

If someone can't use java-library for their API projects then this PR would wreck them.

### DIFF
--- a/changelog/@unreleased/pr-202.v2.yml
+++ b/changelog/@unreleased/pr-202.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: gradle-conjure produces no longer sets up deprecated 'compile' dependencies,
+    instead preferring to declare 'api' dependencies (introduced with the java-library
+    plugin).
+  links:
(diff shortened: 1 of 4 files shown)


carterkozak (approved):
very nice

ferozco:
👍

PR #257, opened 2019-10-07
Don't set source dirs with immutable set

CRogers:
## Before this PR
If anyone tries to modify the IDEA source sets after `gradle-conjure` has set them, like a certain internal gradle plugin, it will fail.

## After this PR
==COMMIT_MSG==
IDEA plugin source sets are no longer set as immutable, so modifications to it by other plugins can occur.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -625,10 +625,12 @@ public final class ConjurePlugin implements Plugin<Project> {
             // module.getSourceDirs / getGeneratedSourceDirs could be an immutable set, so defensively copy
             IdeaModule module = plugin.getModel().getModule();
             module.setSourceDirs(mutableSetWithExtraEntry(
-                    module.getSourceDirs(), project.file(JAVA_GENERATED_SOURCE_DIRNAME)));
+                    module.getSourceDirs(),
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -623,16 +624,11 @@ static void applyDependencyForIdeTasks(Project project, Task compileConjure) {
 
             // module.getSourceDirs / getGeneratedSourceDirs could be an immutable set, so defensively copy
```
CRogers, line 625: This is funny, as we say this this but then make sure it's an immutable set to mess with anyone else who wants to do stuff :p

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -642,6 +638,12 @@ static void applyDependencyForIdeTasks(Project project, Task compileConjure) {
         });
     }
 
+    private static <T> Set<T> mutableSetWithExtraEntry(Set<T> set, T extraItem) {
+        Set<T> newSet = new HashSet<>(set);
```
iamdanfox: can we use LinkedHashSet to prevent future annoying bugs?

iamdanfox (approved):
Would be good to also make alta-client-java tolerant of this being an immutableset

PR #258, opened 2019-10-07
Actually don't set source sets as immutable

CRogers:
## Before this PR
We tried to not set source sets as immtuables in #257, but made a stupid typo

## After this PR
==COMMIT_MSG==
The previous release had a bug, this release fixes that bug and correctly configures source sets as not immutable.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-258.v2.yml
+++ b/changelog/@unreleased/pr-258.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: The previous release had a bug, this release fixes that bug and correctly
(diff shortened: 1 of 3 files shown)


iamdanfox:
Ok so is this the point where we should add a test?

CRogers:
@iamdanfox I actually just removed the code, as it seems to be added automagically by the intellij plugin (I think because we have added it to the main java source set)

CRogers:
There's a test to show this works

iamdanfox (approved):

PR #310, opened 2019-12-03
Set the path sensitivity of ConjureGeneratorTask sources

AlexLandau:
This allows the Gradle build cache to function across different
machines (or different repo locations on the same file system)
when running ConjureGeneratorTasks.

## Before this PR
Remote Gradle build caching works for compileIr tasks, but not e.g. compileConjureObjects tasks.

## After this PR
Remote Gradle build caching works for ConjureGeneratorTasks.

## Possible downsides?
If a custom generator uses the non-relative part of the source path (i.e. the repo location) for something, caching could become inaccurate. This seems unlikely to be something someone would do intentionally. A possible protection against this would be passing a relative path to `outputDirectoryFor` rather than an absolute path. That would protect against `outputDirectoryFor` overrides, though not `compileFiles` overrides.

### DIFF
--- a/changelog/@unreleased/pr-310.v2.yml
+++ b/changelog/@unreleased/pr-310.v2.yml
@@ -1,19 +1,5 @@
 type: improvement
 improvement:
-  description: |-
-    Set the path sensitivity of ConjureGeneratorTask sources
-
-    This allows the Gradle build cache to function across different
-    machines (or different repo locations on the same file system)
-    when running ConjureGeneratorTasks.
-
-    ## Before this PR
-    Remote Gradle build caching works for compileIr tasks, but not e.g. compileConjureObjects tasks.
-
-    ## After this PR
-    Remote Gradle build caching works for ConjureGeneratorTasks.
-
-    ## Possible downsides?
-    If a custom generator uses the non-relative part of the source path (i.e. the repo location) for something, caching could become inaccurate. This seems unlikely to be something someone would do intentionally. A possible protection against this would be passing a relative path to `outputDirectoryFor` rather than an absolute path. That would protect against `outputDirectoryFor` overrides, though not `compileFiles` overrides.
+  description: Enable Gradle remote build caches for ConjureGeneratorTasks.
   links:
   - https://github.com/palantir/gradle-conjure/pull/310
--- a/changelog/@unreleased/pr-310.v2.yml
+++ b/changelog/@unreleased/pr-310.v2.yml
@@ -1,0 +1,19 @@
+type: improvement
+improvement:
+  description: |-
+    Set the path sensitivity of ConjureGeneratorTask sources
+
+    This allows the Gradle build cache to function across different
+    machines (or different repo locations on the same file system)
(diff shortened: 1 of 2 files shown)


changelog/@unreleased/pr-310.v2.yml
```diff
@@ -0,0 +1,19 @@
...
+    when running ConjureGeneratorTasks.
+
+    ## Before this PR
+    Remote Gradle build caching works for compileIr tasks, but not e.g. compileConjureObjects tasks.
+
+    ## After this PR
+    Remote Gradle build caching works for ConjureGeneratorTasks.
+
+    ## Possible downsides?
+    If a custom generator uses the non-relative part of the source path (i.e. the repo location) for something, caching could become inaccurate. This seems unlikely to be something someone would do intentionally. A possible protection against this would be passing a relative path to `outputDirectoryFor` rather than an absolute path. That would protect against `outputDirectoryFor` overrides, though not `compileFiles` overrides.
```
carterkozak: Can we update this to describe the improvement? The "after this PR" section would be sufficient.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java
```diff
@@ -51,6 +52,13 @@ public void execute(Task _task) {
         });
     }
 
+    // Set the path sensitivity of the sources, which would otherwise default to ABSOLUTE
+    @Override
+    @PathSensitive(PathSensitivity.RELATIVE)
+    public FileTree getSource() {
```
AlexLandau: There's a failing ErrorProne check about this being extensible but not having a JavaDoc to explain what to do with it. I'm not sure whether to prefer leaving this extensible so people can override the path sensitivity in subclasses, or just making this final and keeping things simpler.

carterkozak: Everything else is final, this method likely should be as well

carterkozak:
I’ve a couple folks who are more familiar with gradle than myself.

I’d love to have a test that covered this, but it looks like we don’t have much coverage for that sort of thing, only correctness, so I don’t think that’s a blocker.

changelog/@unreleased/pr-310.v2.yml
AlexLandau: Done

iamdanfox:
I've written an example test that exercises the build-cache codepath here: https://github.com/palantir/gradle-conjure/pull/312, but I think testing your PR would require something a little more complicated, as we'd want to verify that _relocating_ your working directory still gets a cache hit.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java
AlexLandau: I made this final.

carterkozak:
👍

PR #312, opened 2019-12-04
build cache test

iamdanfox:
## Before this PR

I added the `@CacheableTask` annotations back in May (https://github.com/palantir/gradle-conjure/pull/141) but only validated them using Gradle Enterprise instead of using tests!

## After this PR
==COMMIT_MSG==
Tests validate that the `compileIr` task results can be loaded from a Gradle Build Cache even when the 'build' directory is wiped.
==COMMIT_MSG==

The PR doesn't actually use their test kit (https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-build-cache) as we're going via nebula here.

### DIFF
--- a/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
+++ b/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
@@ -16,8 +16,10 @@
 
 package com.palantir.gradle.conjure
 
+import java.nio.file.Files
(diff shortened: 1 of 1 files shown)


dansanduleac (approved):

dansanduleac:
Looks good, but FYI appveyor seems to be brekaing with a legit UnusedVariable error coming from errorprone. I wonder why we're not hitting that on the normal CI?

iamdanfox:
I think its a forward slash vs backslash problem - it's been red on develop for ages :/

PR #321, opened 2019-12-12
Add listenablefuture dependency to retrofit project

pkoenig10:
## Before this PR
The retrofit API project does not include a dependency on Guava when using `ListenableFuture`. This will become more problematic once `ListenableFuture` is the default (see palantir/conjure-java#641).

## After this PR
The retrofit API project includes a dependency on Guava.

### DIFF
--- a/changelog/@unreleased/pr-321.v2.yml
+++ b/changelog/@unreleased/pr-321.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Retrofit API projects now include a dependency on Guava.
+  links:
+  - https://github.com/palantir/gradle-conjure/pull/321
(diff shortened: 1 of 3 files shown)


pkoenig10:
@iamdanfox @robert3005 can I get a review?

carterkozak:
I wonder if we should add a dependency on the minimal listenablefuture library instead, guava depends on a “999” versioned empty jar in order to resolve to guava when guava is available.
This way we don’t add a guava dependency to projects which don’t already have one.

carterkozak:
I think I sent you down the wrong path here, sorry! When we pull this into a project with newer guava, the listenablefuture library is upgraded to one that doesn't provide the right dependencies.
I think listenablefuture 999* should depend on guava (circular dependency) to resolve this case automatically, but that's not currently the case.
Sorry for the runaround.

carterkozak:
lgtm, thanks! :+1:

PR #322, opened 2019-12-12
Add validation-api dependency to jersey projects

### LINKED ISSUE #271: automatically add javax validation api dependency if requireNotNullAuthAndBodyParams flag is set
## What happened?
When enabling the `requireNotNullAuthAndBodyParams` flag to generate code with `@NotNull` annotations, in order to actually get validation it's necessary to manually add the `javax.validation:validation-api` dependency manually to the `-jersey` subproject.

## What did you want to happen?
Just like we add conjure-lib, retrofit etc automatically as dependencies in the right places, it'd be if the validation-api dependency was added automatically.

(I tried to take a stab at this, but couldn't figure out the point at which I both have my hands on the generator options but dependencies aren't yet frozen - i.e. my attempt ended in a `org.gradle.api.InvalidUserDataException: cannot change dependencies of configuration ":api:api-jersey:compile" after it has been included in dependency resolution."`)

pkoenig10:
Fixes #271

## Before this PR
Users that want to use `requireNotNullAuthAndBodyParams` have to manually add the `jakarta.validation:jakarta.validation-api` to their Jersey projects.

## After this PR
The `jakarta.validation:jakarta.validation-api` dependency is applied by default to all Jersey projects.

## Possible downsides?
- This dependency is applied even if you aren't using `requireNotNullAuthAndBodyParams`. However the dependency is pretty small and we are planning to make `requireNotNullAuthAndBodyParams` enabled by default (https://github.com/palantir/conjure-java/issues/51).

### DIFF
--- a/changelog/@unreleased/pr-322.v2.yml
+++ b/changelog/@unreleased/pr-322.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Jersey API projects now include a dependency on `jakarta.validation:jakarta.validation-api`.
+  links:
+  - https://github.com/palantir/gradle-conjure/pull/322
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -341,6 +341,7 @@ public final class ConjurePlugin implements Plugin<Project> {
                 subproj.getDependencies().add("api", project.findProject(objectsProjectName));
                 subproj.getDependencies().add("api", "javax.ws.rs:javax.ws.rs-api");
                 subproj.getDependencies().add("compileOnly", ANNOTATION_API);
+                subproj.getDependencies().add("implementation", "jakarta.validation:jakarta.validation-api");
             });
         }
     }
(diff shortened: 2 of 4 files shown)


ferozco:
I feel like we should avoid automatically adding dependencies especially if they are only needed when a certain feature flag is set. Perhaps we can improve the documentation in conjure-java so that it is clear what users need to do when setting the feature flag.

pkoenig10:
What's the downside of adding this dependency? It's a pretty small dependency that users are very likely pulling in already from another dependency.

In this case I think the bad UX of having to manually add dependencies with feature flags outweighs the cost of the added dependency.

ferozco:
I'd like to avoid adding the dependencies so that gradle-conjure knows as little as possible about conjure-java codegen and so that users that don't use the feature flag do not need to add unnecessary constraints to their versions.props.

This is only an issue for existing repos that adopt the flag since all new repos (at least internally) come with the feature flag and the dependency. Hence, the suggestion to update the docs in conjure-java so people know to do the right thing when changing their codegen

pkoenig10:
I've updated this PR to only add the dependency if the `requireNotNullAuthAndBodyParams` option is set.

pkoenig10:
@ferozco any other comments?

ferozco:
@pkoenig10 as I mentioned before gradle-conjure should be as a decoupled from the generators as possible to prevent any inadvertent breaks as well as avoid requiring lockstep upgrades between the generator and the build tool. I don't think we should go down this route and instead think we should add improve the documentation in conjure-java so that users who adopt a feature flag know how to solve the any compilation breaks that it may cause

robert3005:
@ferozco I think something is backward here. Why does this issue exist only in java but not in case of typescript for instance? Should we reconsider how we integrate gradle with conjure such that generators can output appropriate build files instead of relying on users doing so?

ferozco:
@robert3005 I totally agree it would be cool if conjure-java could emit a definition of what was required to build the code it generates. A while back @dfox and I spent sometime brainstorming solutions but ended up deciding against investing in it since it didn't seem worth the cost. I still agree with that conclusion and don't think we should invest time in solving this particular problem at this time

PR #331, opened 2019-12-19
Specify a version for annotation-api

iamdanfox:
## Before this PR

We're starting to move from javax -> jakarta as a consequence of taking a [jersey upgrade in c-j-r](https://github.com/palantir/conjure-java-runtime/pull/1372) (and transitively in WC).  Unfortunately, this means that the wc-bom has suddently stopped providing a version recommendation of `javax.annotation:javax.annotation-api = 1.3.2`.  This means people's WC upgrades are getting stuck with errors like the following.

```
Could not determine the dependencies of task ':rescue:compileJava'.
> Could not resolve all task dependencies for configuration ':rescue-internal-objects:compileClasspath'.
   > Could not find javax.validation:validation-api:.
     Required by:
         project :rescue-internal-objects
```

Note that this happens even with my gradle-jakarta-renames plugin applied, as the resolution comes later.

## After this PR
==COMMIT_MSG==
gradle-conjure no longer requires users to provide a version number for `javax.annotation:javax.annotation-api`, instead it will just set up a dependency on latest (1.3.2).  This can still be substituted later with the new jakarta version.
==COMMIT_MSG==

I considered switching to jakata straight away here (https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api) but figured it's best to do minimal changes to unblock things first, as I really want to get baseline-class-uniqueness running everywhere before we do more of this.

## Possible downsides?
- I honestly can't remember why we didn't do this originally... it seems to just be net better?

### DIFF
--- a/changelog/@unreleased/pr-331.v2.yml
+++ b/changelog/@unreleased/pr-331.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: gradle-conjure no longer requires users to provide a version number
+    for `javax.annotation:javax.annotation-api`, instead it will just set up a dependency
+    on latest (1.3.2).  This can still be substituted later with the new jakarta version.
+  links:
+  - https://github.com/palantir/gradle-conjure/pull/331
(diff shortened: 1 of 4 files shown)


ferozco:
In general I'm hesitant to have the plugin provide a version for the dependency since it makes it more confusing for users. I think the ideal way to handle this would be to use gradle capabilities to allow you to have both dependencies and pick one of them.

iamdanfox:
> Note that this happens even with my gradle-jakarta-renames plugin applied, as the resolution comes later.

My plugin already sets up capabilities but we don't seem to get to that stage.

I don't really understand how this is "more confusing for users" though?

ferozco:
Its just unclear where the version is coming from since its baked into the plugin code. As an alternative, We could excavate out an addition into the versions.props files of users as part of the upgrade and then have baseline clean it up later once we've moved over to jakarta

iamdanfox:
I'm not convinced that anyone will ever ping our help channel with that complaint though? We already bake real dependencies into plugin code, and people seem to manage with these...

ferozco:
ok, lets move forward 👍

PR #363, opened 2020-02-04
add conjure-java-local plugin

ferozco:
## Before this PR
We had some support for java local codegen, but it definitely wasn't supported in a first class way.

## After this PR
==COMMIT_MSG==
Add new `conjure-java-local` plugin which allows for first class Java local codegen
==COMMIT_MSG==

This plugin will generate bindings for each individual API in child project. Below an example of what the expected project structure would look like:
directories:
```
└── remote-api/
    ├── bar-api/
    ├── build.gradle
    └── foo-api/
```

build.gradle
```gradle
apply plugin: 'conjure-java-local'

conjure {
  java {
    addFlag 'objects'
    addFlag 'jersey'
    addFlag 'strictObjects'
  }
}

dependencies {
   conjure 'com.palantir.foo:foo-api:1.0.0'
   conjure 'com.palantir.bar:bar-api:2.0.0'
}
```

## Possible downsides?
The plugin is incredibly unopinionated, only requiring that a package prefix is provided (defaulting to the current project group), so users are able to locally codegen anything that they would like.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -37,7 +37,7 @@ import org.gradle.api.tasks.TaskProvider;
 public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
     private static final String CONJURE_CONFIGURATION = "conjure";
     private static final Pattern DEFINITION_NAME =
-            Pattern.compile("(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?)(.conjure)?.json");
+            Pattern.compile("(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?)(\\.conjure)?\\.json");
 
     static final String CONJURE_JAVA = "conjureJava";
     static final String CONJURE_JAVA_BINARY = "com.palantir.conjure.java:conjure-java";
@@ -49,7 +49,7 @@ public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
         ConjureExtension extension =
                 project.getExtensions().create(ConjureExtension.EXTENSION_NAME, ConjureExtension.class);
 
-        Configuration conjureIrConfiguration = project.getConfigurations().maybeCreate(CONJURE_CONFIGURATION);
+        Configuration conjureIrConfiguration = project.getConfigurations().create(CONJURE_CONFIGURATION);
         TaskProvider<Copy> extractConjureIr = project.getTasks().register("extractConjureIr", Copy.class, task -> {
             task.rename(DEFINITION_NAME, "$1.conjure.json");
             task.from(conjureIrConfiguration);
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -75,7 +75,7 @@ public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
         // Validating that each subproject has a corresponding definition and vice versa.
         // We do this in afterEvaluate to ensure the configuration is populated.
         project.afterEvaluate(p -> {
-            Set<String> apis = conjureIrConfiguration.getAllDependencies().stream()
+            Set<String> apis = conjureIrConfiguration.getIncoming().getDependencies().stream()
                     .map(Dependency::getName)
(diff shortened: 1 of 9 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.plugins.BasePlugin;
+import org.gradle.api.plugins.JavaLibraryPlugin;
+import org.gradle.api.tasks.Copy;
+import org.gradle.api.tasks.TaskProvider;
+
+public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
+    private static final String CONJURE_CONFIGURATION = "conjure";
+    private static final Pattern DEFINITION_NAME =
+            Pattern.compile("(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?)(.conjure)?.json");
```
dansanduleac: might wanna `\\.conjure` and `\\.json`

ferozco: 🤦‍♂ good catch

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+    static final String CONJURE_JAVA_BINARY = "com.palantir.conjure.java:conjure-java";
+
+    @Override
+    public void apply(Project project) {
+        project.getPlugins().apply(BasePlugin.class);
+
+        ConjureExtension extension =
+                project.getExtensions().create(ConjureExtension.EXTENSION_NAME, ConjureExtension.class);
+
+        Configuration conjureIrConfiguration = project.getConfigurations().maybeCreate(CONJURE_CONFIGURATION);
```
dansanduleac: I prefer failing here if it already exists, there's no good reason why the user would roll their own and only undefined behaviour can come out of that if it were allowed

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
dansanduleac: yeah just cos some people might unconsciously write -conjure or whatever :D

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+                project.getExtensions().create(ConjureExtension.EXTENSION_NAME, ConjureExtension.class);
+
+        Configuration conjureIrConfiguration = project.getConfigurations().create(CONJURE_CONFIGURATION);
+        TaskProvider<Copy> extractConjureIr = project.getTasks().register("extractConjureIr", Copy.class, task -> {
+            task.rename(DEFINITION_NAME, "$1.conjure.json");
+            task.from(conjureIrConfiguration);
+            task.into(project.getLayout().getBuildDirectory().dir("conjure-ir"));
+        });
+
+        Configuration conjureJavaConfig = project.getConfigurations().maybeCreate(CONJURE_JAVA);
```
dansanduleac: create here too

dansanduleac:
```suggestion
        Configuration conjureJavaConfig = project.getConfigurations().create(CONJURE_JAVA);
```

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+            Project project,
+            ConjureExtension extension,
+            ExtractExecutableTask extractJavaTask,
+            TaskProvider<Copy> extractConjureIr,
+            Configuration conjureIrConfiguration) {
+
+        // Validating that each subproject has a corresponding definition and vice versa.
+        // We do this in afterEvaluate to ensure the configuration is populated.
+        project.afterEvaluate(p -> {
+            Set<String> apis = conjureIrConfiguration.getAllDependencies().stream()
```
dansanduleac: `getIncoming().getDependencies()` as it ensures that dependency actions are run too (e.g. default dependencies, and actions registered via `withDependencies`)

dansanduleac:
```suggestion
            Set<String> apis = conjureIrConfiguration.getIncoming().getDependencies().stream()
```

iamdanfox (approved):

PR #371, opened 2020-02-12
Fix: Use jakarta-api instead javax-api

robert3005:
## Before this PR
We had been using old javax.ws.rs:javax.ws.rs-api which can lead to build failures via checkUnusedDependencies from gradle-baseline

## After this PR
==COMMIT_MSG==
Stop using legacy javax.ws.rs:javax.ws.rs-api when compiling jersey projects
==COMMIT_MSG==

## Possible downsides?
N/A, this api should be backwards compatible


ferozco:
👍

PR #372, opened 2020-02-12
Remove appveyor

ferozco:
We weren't blocking PRs on it so it was just introducing noise


iamdanfox (approved):
Also CIrcleCI can do windows CI now, so if we re-enable this we can do it all within Circle

PR #377, opened 2020-02-14
add first class support for dialogue

ferozco:
## Before this PR

To be able to use locally defined github YML in pdeps-bot.

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -182,40 +182,41 @@ public final class ConjurePlugin implements Plugin<Project> {
             ConjureProductDependenciesExtension productDependencyExt,
             ExtractExecutableTask extractJavaTask) {
         String dialogueProjectName = project.getName() + JAVA_DIALOGUE_SUFFIX;
-        if (project.findProject(dialogueProjectName) != null) {
-            String objectsProjectName = project.getName() + JAVA_OBJECTS_SUFFIX;
-            if (project.findProject(objectsProjectName) == null) {
-                throw new IllegalStateException(
-                        String.format("Cannot enable '%s' without '%s'", dialogueProjectName, objectsProjectName));
-            }
-
-            project.project(dialogueProjectName, subproj -> {
-                subproj.getPluginManager().apply(JavaLibraryPlugin.class);
-                ignoreFromCheckUnusedDependencies(subproj);
-                addGeneratedToMainSourceSet(subproj);
-                project.getTasks().create("compileConjureDialogue", ConjureGeneratorTask.class, task -> {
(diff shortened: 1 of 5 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -169,6 +169,52 @@ private static void setupConjureJavaProject(
...
+
+    private static void setupConjureDialogueProject(
+            Project project,
+            Supplier<GeneratorOptions> optionsSupplier,
+            Task compileConjure,
+            Task compileIrTask,
+            ConjureProductDependenciesExtension productDependencyExt,
+            ExtractExecutableTask extractJavaTask) {
+        String dialogueProjectName = project.getName() + JAVA_DIALOGUE_SUFFIX;
+        if (project.findProject(dialogueProjectName) != null) {
```
iamdanfox: early returns over nesting?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -169,6 +169,52 @@ private static void setupConjureJavaProject(
...
+            String objectsProjectName = project.getName() + JAVA_OBJECTS_SUFFIX;
+            if (project.findProject(objectsProjectName) == null) {
+                throw new IllegalStateException(
```
iamdanfox: log.warn instead maybe? if people's service definitions don't actually include _any_ conjure objects?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -169,6 +169,52 @@ private static void setupConjureJavaProject(
...
+                    task.dependsOn(createWriteGitignoreTask(
+                            subproj, "gitignoreConjureDialogue", subproj.getProjectDir(), JAVA_GITIGNORE_CONTENTS));
+                    task.dependsOn(extractJavaTask);
+                });
+
+                ConjureJavaServiceDependencies.configureJavaServiceDependencies(subproj, productDependencyExt);
+                Task cleanTask = project.getTasks().findByName(TASK_CLEAN);
+                cleanTask.dependsOn(project.getTasks().findByName("cleanCompileConjureDialogue"));
+                subproj.getDependencies().add("api", project.findProject(objectsProjectName));
+                subproj.getDependencies().add("api", "com.palantir.dialogue:dialogue-target");
```
iamdanfox: think the project structure is gonna be _mildly_ annoying if we want to make atomic changes across the codegen & target lirbary at the same time, as users will have to lock-step upgrade two deps or get confusnig compilation failures from generated code

iamdanfox:
Could we hold this PR until we get a release of conjure-java that includes dialogue so we can do an actual test here?

carterkozak (approved):

PR #381, opened 2020-02-19
plumb product dependencies to conjure core

ferozco:
## Before this PR
We only passed product dependency information to the the generators.

## After this PR
Following https://github.com/palantir/conjure/pull/532 we can now add the product dependency information to the IR itself. We'll probably want to update the generators (see Conjure-TS) to take advantage of the extension if present.

We may also want to consider adding other information like name, version, source repository, etc. into the extension for use by generators like Conjure-Java dialogue and Conjure-python
==COMMIT_MSG==
Add product dependency information into produced Conjure IR
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -17,11 +17,16 @@
 package com.palantir.gradle.conjure;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.palantir.gradle.conjure.api.ServiceDependency;
 import java.io.File;
+import java.io.IOException;
 import java.util.List;
 import java.util.function.Supplier;
 import org.gradle.api.DefaultTask;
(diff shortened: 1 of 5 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -73,11 +85,22 @@ public final void generate() {
                     new File(executableDir.get(), EXECUTABLE).getAbsolutePath(),
                     "compile",
                     inputDirectory.get().getAbsolutePath(),
-                    outputFile.getAbsolutePath());
+                    outputFile.getAbsolutePath(),
+                    "--extensions",
```
CRogers: Is this extensions flag some pre-existing thing (is it part of the spec?)

ferozco: this is something that we are adding here: https://github.com/palantir/conjure/pull/532

CRogers (approved):

PR #384, opened 2020-02-21
Ensure stderr of conjure IR run is printed it in exception method when it fails

CRogers:
## Before this PR
The output of the conjure IR task is not included in the exception when it fails, so you end up with a very not-useful error message:

Screen Shot 2020-02-21 at 2 27 38 PM

## After this PR
==COMMIT_MSG==
Error messages produced by conjure IR compilation will not appear in the exception, so at the top of circle builds.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -17,6 +17,7 @@
 package com.palantir.gradle.conjure;
 
 import com.google.common.collect.ImmutableList;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.util.List;
 import java.util.function.Supplier;
@@ -27,6 +28,7 @@ import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.PathSensitivity;
 import org.gradle.api.tasks.TaskAction;
+import org.gradle.process.ExecResult;
 
 @CacheableTask
 public class CompileIrTask extends DefaultTask {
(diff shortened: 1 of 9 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -67,17 +69,28 @@ public final File getExecutableDir() {
...
+                inputDirectory.get().getAbsolutePath(),
+                outputFile.getAbsolutePath());
+
+        List<String> args = commandArgsBuilder.build();
+
+        ExecResult execResult = getProject().exec(execSpec -> {
             getLogger().info("Running compiler with args: {}", args);
             execSpec.commandLine(args.toArray());
+            execSpec.isIgnoreExitValue();
+            execSpec.setStandardOutput(output);
```
carterkozak: Can we capture stderr here as well?

CRogers: This is both stdout and stderr

carterkozak: ha overloading :-) Thanks

ferozco (commented):
Can we do this for all of the generators while we're at it?

carterkozak:
Any way we can write a test for this?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
CRogers: Was confused, this is actually just stdout like you thought. Now I'm just doing stderr.

carterkozak: Any reason not to write both to our output buffer?

CRogers:
Ok, I think I've fixed all your comments. Now with tests and for all the generators.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -0,0 +1,46 @@
...
+            project.getLogger().info("Running with args: {}", args);
+            execSpec.commandLine(args.toArray());
+            execSpec.setIgnoreExitValue(true);
+            execSpec.setStandardOutput(output);
+            execSpec.setErrorOutput(output);
+        });
+
+        if (execResult.getExitValue() != 0) {
+            throw new RuntimeException(String.format(
+                    "Failed to " + failedTo + ". The command '%s' failed with exit code %d. Output:\n%s",
```
carterkozak: constant format strings, otherwise a `%s` in the `failedTo` string will offset args
```suggestion
                    "Failed to %s. The command '%s' failed with exit code %d. Output:\n%s", failedTo
```

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -0,0 +1,46 @@
...
+
+    static void exec(Project project, String failedTo, List<String> args) {
+        ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+        ExecResult execResult = project.exec(execSpec -> {
+            project.getLogger().info("Running with args: {}", args);
+            execSpec.commandLine(args.toArray());
+            execSpec.setIgnoreExitValue(true);
+            execSpec.setStandardOutput(output);
+            execSpec.setErrorOutput(output);
```
carterkozak, line 41: thanks!

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
CRogers: Nice catch!

carterkozak:
lgtm, thanks! :+1:

PR #385, opened 2020-02-21
Inject product dependencies into local generate java projects

ferozco:
## Before this PR
We did not embed product dependencies into locally generated Java projects

## After this PR
==COMMIT_MSG==
Embed product dependencies in locally generated Java projects using `"recommended-product-dependencies"` in Conjure IR extensions
==COMMIT_MSG==

## Possible downsides?
N/A


gradle-conjure/build.gradle
```diff
@@ -24,7 +24,8 @@ dependencies {
     compile project(':gradle-conjure-api')
     compile 'com.google.guava:guava'
     compile 'com.palantir.sls.versions:sls-versions'
-    compile 'com.fasterxml.jackson.core:jackson-databind'
+    compile 'com.palantir.conjure.java.runtime:conjure-java-jackson-serialization'
+    compile 'com.palantir.sls-packaging:gradle-recommended-product-dependencies'
```
iamdanfox: might as well use `implementation` for these? I know nobody should be compiling against these, but worth sticking to best practises?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -134,4 +150,17 @@ private static void createGenerateTask(
...
+    private static Set<ProductDependency> extractProductDependencies(File irFile) {
+        try {
+            MinimalConjureDefinition conjureDefinition =
+                    OBJECT_MAPPER.readValue(irFile, MinimalConjureDefinition.class);
+            return conjureDefinition
+                    .extensions()
+                    .map(MinimalConjureDefinition.Extensions::productDependencies)
+                    .orElseGet(Collections::emptySet);
+        } catch (IOException e) {
+            throw new SafeRuntimeException("Failed to parse conjure definition", e);
```
iamdanfox, line 163: wasn't the idea that we should be able to limp onwards whatever the block contains?

gradle-conjure/build.gradle
```diff
@@ -24,7 +24,8 @@ dependencies {
     compile project(':gradle-conjure-api')
     compile 'com.google.guava:guava'
     compile 'com.palantir.sls.versions:sls-versions'
-    compile 'com.fasterxml.jackson.core:jackson-databind'
+    compile 'com.palantir.conjure.java.runtime:conjure-java-jackson-serialization'
```
iamdanfox: do we _really_ need all this? jackson afterburner etc?

ferozco: I just wanted access to a consistent well configured object mapper, wasn't overly concerned about all the dependencies

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
ferozco: We fail if the block exists and it contains a `recommended-product-dependencies` field which has something thing that doesn't look like product dependencies. I thought it was preferable to fail in this case rather than silently ignore malformed pdeps

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPluginIntegrationSpec.groovy
```diff
@@ -139,6 +141,53 @@ class ConjureJavaLocalCodegenPluginIntegrationSpec extends IntegrationSpec {
...
+
+        when:
+        ExecutionResult result = runTasksSuccessfully('jar')
+
+        then:
+        result.wasExecuted(':conjure-api:compileJava')
+        result.wasExecuted(':conjure-api:generateConjure')
+
+        def recommendedProductDependencies = readRecommendedProductDeps(file('conjure-api/build/libs/conjure-api-1.0.0.jar'))
+        recommendedProductDependencies == '{"recommended-product-dependencies":[{' +
```
iamdanfox, line 182: weren't we gonna call this `sls-recommended-product-dependencies` for consistency with what we embed in the jar manifest?

ferozco: The Jar manifest looks like: 
```
Sls-Recommended-Product-Dependencies: {"recommended-product-dependencies
 ":[{"product-group":"com.palantir.food","product-name":"foo-server","mi
 nimum-version":"3.9.0","maximum-version":"3.x.x","recommended-version":
 "3.9.0"}]}
```

So I made it `recommended-predocut-dependencies` consistent with the inner json

ferozco: @iamdanfox We haven't released the change yet so we can make adjustments, but this seemed like a a nice level of consistency

iamdanfox (approved):

PR #373, opened 2020-02-12
Ignore unused conjure-lib from java subprojects

dansanduleac:
## Before this PR

Users of baseline's [checkUnusedDependencies](https://github.com/palantir/gradle-baseline/#compalantirbaseline-exact-dependencies) get an error if they enable it on conjure java subprojects:

```
> Found 2 dependencies unused during compilation, please delete them from 'foo/foo-objects/build.gradle' or choose one of the suggested fixes:
        com.palantir.conjure.java:conjure-lib
                Did you mean:
                        implementation com.fasterxml.jackson.core:jackson-annotations
                        implementation com.fasterxml.jackson.core:jackson-databind
                        implementation com.palantir.ri:resource-identifier
                        implementation com.palantir.safe-logging:preconditions
                        implementation com.palantir.safe-logging:safe-logging
        javax.ws.rs:javax.ws.rs-api
```

## After this PR
==COMMIT_MSG==
If `com.palantir.baseline-exact-dependencies` is found, configure its task to ignore `conjure-lib` because it is always added to subprojects by this plugin.
==COMMIT_MSG==

PR #393, opened 2020-02-28
Configure checkUnusedDependencies by task type

dansanduleac:
## Before this PR

With https://github.com/palantir/gradle-baseline/pull/1262 the existing logic here from #373 will start failing.

## After this PR
==COMMIT_MSG==
Use a more robust way to configure the `checkUnusedDependencies` tasks for conjure projects.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -391,18 +391,30 @@ public final class ConjurePlugin implements Plugin<Project> {
         }
     }
 
+    @SuppressWarnings("unchecked")
     private static void ignoreFromCheckUnusedDependencies(Project proj) {
-        proj.getPluginManager().withPlugin("com.palantir.baseline-exact-dependencies", plugin -> {
-            Object task = proj.getTasks().getByName("checkUnusedDependencies");
+        proj.getPlugins().withId("com.palantir.baseline-exact-dependencies", plugin -> {
+            Class<? extends Task> checkUnusedDependenciesTask;
             try {
-                Method ignoreMethod = task.getClass().getMethod("ignore", String.class, String.class);
-                List<String> conjureJavaLibComponents = Splitter.on(':').splitToList(CONJURE_JAVA_LIB_DEP);
-                ignoreMethod.invoke(task, conjureJavaLibComponents.get(0), conjureJavaLibComponents.get(1));
-                // also ignore guava since retrofit adds it...
-                ignoreMethod.invoke(task, "com.google.guava", "guava");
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -391,18 +391,30 @@ private static void setupConjureUndertowProject(
...
+        proj.getPlugins().withId("com.palantir.baseline-exact-dependencies", plugin -> {
+            Class<? extends Task> checkUnusedDependenciesTask;
             try {
-                Method ignoreMethod = task.getClass().getMethod("ignore", String.class, String.class);
-                List<String> conjureJavaLibComponents = Splitter.on(':').splitToList(CONJURE_JAVA_LIB_DEP);
-                ignoreMethod.invoke(task, conjureJavaLibComponents.get(0), conjureJavaLibComponents.get(1));
-                // also ignore guava since retrofit adds it...
-                ignoreMethod.invoke(task, "com.google.guava", "guava");
-            } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+                ClassLoader baselineClassloader = plugin.getClass().getClassLoader();
```
dansanduleac, line 399: this is a neat way to get the right classloader, whether baseline is applied via `plugins{}` or buildscript.

ferozco: Why not just disable the entire task on conjure subprojects?

ferozco:
This is some pretty gnarly code. Could we add a compileOnly dependency on baseline to avoid the dynamic method invocation?

dansanduleac:
I don't think that works, I've tried that before and ran into the problem that as soon as you have a class that deals with external compileOnly stuff, you can _only_ load that class in your _current_ classloader (even if you load it via reflection), which will then not be able to access the baseline classes from another classloader.

Only way around that is to play with UrlClassloaders and splice them in a way that's convenient, such that the *lowest level* (farthest from the parent) classloader _also_ happens to contain baseline's classloader as a child.

e.g.

```
ROOT
\- our current classloader (contains gradle-conjure and deps)
\- baseline plugin classloader
\- manufactured classloader which is a copy of our current classloader
   \- baseline plugin classloader
```

This is a consequence of how Java loads classes - given a classloader and a class to load, it will choose the first classloader going down from the root, and so whatever your loaded class references must all exist between the root and that classloader.

CRogers (approved):

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
dansanduleac: Have considered this too, but ultimately I prefer to still restrict the jars people depend on, especially for api modules, which are supposed to be quite light.

PR #399, opened 2020-03-04
Replace javax.annotation-api with jakarta.annoatation-api

ferozco:
## After this PR
==COMMIT_MSG==
Replace `javax.annotation-api` with `jakarta.annotation-api` as default dependency on 
==COMMIT_MSG==


carterkozak (approved):
Thanks!

PR #417, opened 2020-03-23
Produce `rawIr` task

ferozco:
## Before this PR
It was not possible to get the Conjure IR without resolving the product dependencies, making automatic verification of product dependencies difficult. 

## After this PR
==COMMIT_MSG==
Expose `rawIr` task which produces the raw IR without any extensions.
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -66,6 +66,7 @@ public class CompileIrTask extends DefaultTask {
     }
 
     @OutputFile
+    @PathSensitive(PathSensitivity.NONE)
     public final RegularFileProperty getOutputIrFile() {
         return outputIrFile;
     }
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.function.Supplier;
 import org.gradle.api.DefaultTask;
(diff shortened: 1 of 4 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -37,24 +38,40 @@
...
 
-    public final void setInputDirectory(Supplier<File> inputDirectory) {
-        this.inputDirectory = inputDirectory;
+    /**
+     * Eagerly get where to output the generated IR.
+     *
+     * @deprecated Use {@link #getOutputIrFile()} instead.
+     */
+    @Deprecated
+    public final File getOutputFile() {
```
dansanduleac, line 64: think you want `@Internal` too or else gradle 6 complains

ferozco: I think it only yells at methods with Property return types

dansanduleac: oh i thought it's any method that starts with `get` i.e. that looks like a bean accessor

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -37,24 +38,40 @@
...
+        return outputIrFile.getAsFile().get();
     }
 
     @OutputFile
-    public final File getOutputFile() {
-        return outputFile;
+    public final RegularFileProperty getOutputIrFile() {
```
dansanduleac, line 70: can we put a `@PathSensitivity(NONE)` since the task doesn't care about the path?

dansanduleac (approved):

PR #419, opened 2020-03-23
Fix bad advice

dansanduleac:
## Before this PR

Made it throw a deprecation  warning in gradle 6 in #417

## After this PR
==COMMIT_MSG==
Fix deprecation warning with gradle 6 for `rawIr` and `compileIr`.
==COMMIT_MSG==


ferozco:
👍

PR #420, opened 2020-03-25
Specify an explicit com.palantir.dialogue:dialogue-target version

iamdanfox:
## Before this PR

We have excavators going out to try and get everyone publishing dialogue bindings (roomba/publish-dialogue). Some of these are failing with the following:

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':external-catalog-api:external-catalog-api-dialogue:compileJava'.
> Could not resolve all files for configuration ':external-catalog-api:external-catalog-api-dialogue:compileClasspath'.
   > Could not find com.palantir.dialogue:dialogue-target:.
     Required by:
         project :external-catalog-api:external-catalog-api-dialogue

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

https://internal-circle/gh/workshop/external-catalog/10389

This could be fixed by just upgrading WC, but it seems like these concerns are actually independent and could also be fixed here in gradle-conjure.

## After this PR
==COMMIT_MSG==
Generated -dialogue projects now depend on dialogue-target 1.45.0 by default rather than requiring users to supply a version. Users can still upgrade this if they choose in versions.props.
==COMMIT_MSG==


ferozco:
👍

PR #426, opened 2020-03-27
Install NPM dependencies with the --no-production flag

walkerburgin:
## Before this PR
Setting `NODE_ENV=production` causes dev dependencies such as TypeScript to not be installed. This causes the `compileTypeScript` task to fail with `tsc: not found`  

## After this PR
==COMMIT_MSG==
Always specify the `--no-production` flag when installing dependencies to ignore `NODE_ENV`. For our use case we always want to install dev dependencies. 
==COMMIT_MSG==

## Possible downsides?
None that I can think of.


ferozco:
Wow thats pretty lame. Thanks for the fix 👍

PR #462, opened 2020-05-21
Dialogue projects now depend on 1.50.0

iamdanfox:
Just picking up this https://github.com/palantir/dialogue/pull/784


ferozco:
👍

PR #549, opened 2020-07-31
Pull out IR compilation into separate base plugin

ferozco:
## Before this PR
We had a monolithic plugin that contained all of the logic for going from conjure source to generated code

## After this PR
==COMMIT_MSG==
pull out IR compilation into separate base plugin
==COMMIT_MSG==

## Possible downsides?
N/A


robert3005:
👍

PR #559, opened 2020-08-17
expose conjureIr via consumable configuration

mrauen:
## Before this PR
Internal infrastructure needs to hook into the IRs produced for a subset of the Java projects in a repository. Once the relevant Java projects are specified, this will allow that infrastructure to grab the relevant IRs.

## After this PR

==COMMIT_MSG==
Expose the IRs produced by the compileIr task via a consumable configuration.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
@@ -94,7 +94,7 @@ public final class ConjureBasePlugin implements Plugin<Project> {
             compileIr.getProductDependencies().set(project.provider(pdepsExtension::getProductDependencies));
             compileIr.dependsOn(copyConjureSourcesTask);
             compileIr.dependsOn(extractCompilerTask);
-            
+
             project.getConfigurations().create(CONJURE_IR_CONFIGURATION, conf -> {
                 conf.setCanBeResolved(false);
             });
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
@@ -24,6 +24,7 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.attributes.Attribute;
 import org.gradle.api.file.Directory;
 import org.gradle.api.file.SourceDirectorySet;
 import org.gradle.api.plugins.BasePlugin;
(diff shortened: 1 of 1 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
```diff
@@ -93,6 +94,11 @@ private static void createIrTasks(
             compileIr.getProductDependencies().set(project.provider(pdepsExtension::getProductDependencies));
             compileIr.dependsOn(copyConjureSourcesTask);
             compileIr.dependsOn(extractCompilerTask);
+            
+            project.getConfigurations().maybeCreate(CONJURE_IR_CONFIGURATION);
```
dansanduleac: please use `create`, it should be an error if the user created this configuration already, if the plugin is supposed to own it

dansanduleac: also need to make it non resolvable:
```java
project.getConfigurations().create(CONJURE_IR_CONFIGURATION, conf -> {
    conf.setCanBeResolved(false);
});
```

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
```diff
@@ -93,6 +94,11 @@ private static void createIrTasks(
...
+            project.artifacts(artifactHandler -> {
+                artifactHandler.add(CONJURE_IR_CONFIGURATION, compileIr.getOutputIrFile());
```
dansanduleac: make sure this knows the `compileIr` task need to run to produce this. I think you need an extra param like
```
artifact -> artifact.builtBy(compileIr)
```

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
```diff
@@ -93,6 +94,14 @@ private static void createIrTasks(
             compileIr.getProductDependencies().set(project.provider(pdepsExtension::getProductDependencies));
             compileIr.dependsOn(copyConjureSourcesTask);
             compileIr.dependsOn(extractCompilerTask);
+
+            project.getConfigurations().create(CONJURE_IR_CONFIGURATION, conf -> {
```
ferozco, line 101: You'll want to tag this configuration with some attributes so that gradle selects the right variant in cases where there are multiple outgoing configurations (i.e. the conjure project also has the Java plugin applied)

ferozco: This also means that you'll need to tag the consuming configuration with the same attributes

mrauen:
@ferozco I've added an attribute to the configuration, is there anything outstanding before this can merge?

PR #617, opened 2020-10-12
fix: Use custom Usage on outgoing conjure configuration

ferozco:
## Before this PR
We overrode the gradle default `Usage` attribute for the outgoing configuration that contained the conjureIr. While this seemed nice at first, it actually broke platforms because of how gradles resolution works.

## After this PR
==COMMIT_MSG==
Conjure IR is now exposed on a configuration tagged with attribute of type Usage with name `com.palantir.conjure`
==COMMIT_MSG==

## Possible downsides?
This will break anyone who relied on the previous usage

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
@@ -24,6 +24,7 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.Usage;
 import org.gradle.api.file.Directory;
 import org.gradle.api.file.SourceDirectorySet;
@@ -35,6 +36,8 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin;
 import org.gradle.util.GFileUtils;
 
 public final class ConjureBasePlugin implements Plugin<Project> {
+    static final Attribute<Usage> CONJURE_USAGE = Attribute.of("com.palantir.conjure", Usage.class);
+
     static final String COMPILE_IR_TASK = "compileIr";
     static final String SERVICE_DEPENDENCIES_TASK = "generateConjureServiceDependencies";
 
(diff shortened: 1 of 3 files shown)


carterkozak (approved):
Looks reasonable

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
```diff
@@ -105,8 +108,7 @@ private static void createOutgoingConfiguration(Project project, TaskProvider<Co
             conf.setCanBeResolved(false);
             conf.setCanBeConsumed(true);
             conf.setVisible(true);
-            conf.getAttributes()
-                    .attribute(Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, "conjure"));
+            conf.getAttributes().attribute(CONJURE_USAGE, project.getObjects().named(Usage.class, "conjure"));
```
dansanduleac, line 111: Note that if you specify a specific attribute on the consumer side, gradle considers both configurations that match it, but also configurations that don't have that attribute defined at all.
Therefore, if there are multiple published configurations (i.e. having attributes), some specifying this attribute and others not specifying it, gradle will say they are ambiguous on the consumer side, unless you also specify a disambiguation rule.

I think the previous thing made sense because:
1. a different usage makes it impossible to depend on both the java variant and the conjure variant _in the same configuration_ (which is probably what you want)
2. it doesn't run into the problem above

dansanduleac:
👎 I think this won't do what you want..

> it actually broke platforms because of how gradles resolution works.

maybe I'm not getting the actual use case, so can you explain the above in more detail?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
ferozco: @dansanduleac i've updated the test cases to demonstrate that the issue that you raised doesn't seem to occur here: [`c86317c` (#617)](https://github.com/palantir/gradle-conjure/pull/617/commits/c86317c4b5c306ea16fec0fe3ad06d47e163754c)

dansanduleac:
👍  Thanks for the demonstration, I'm convinced it works in the normal case now.
As discussed on slack, it's also interesting to note that a consuming configuration without any attributes will prefer the java variant rather than failing to disambiguate. It seems this is because of some project-wide disambiguation rules that java-base adds when applied to a project: https://github.com/gradle/gradle/blob/96ecb346486d26297614c48d745473a370a2d68b/subprojects/dependency-management/src/main/java/org/gradle/internal/component/external/model/PreferJavaRuntimeVariant.java#L82-L87

PR #618, opened 2020-10-13
[go-fast] run java generators _in process_ to avoid JVM startup cost

iamdanfox:
## Before this PR

Gotham is a very large project, with many conjure projects. AFAIK the startup cost of so many conjure generators has become annoying. This is incurred because of the stance we took originally in RFC 002, with the intention of defining a clear separation between build tools (like gradle-conjure) and generators like conjure-java, in anticipation of massive community conjure adoption (lol).

There is a PR up to relax our RFC002 stance and just publish a jar for conjure-java: https://github.com/palantir/conjure-java/pull/1112/files.  This would then be incorporated into a gotham-custom code generation plugin. To achieve full gains, this same change would likely need to be made to the conjure CLI itself (and possibly conjure-python for completeness).  Then, we'd either need to adopt this same "depend on jars" codepath in gradle-conjure.  Also, we'd need to be careful about the ABI/API of conjure-java classes to ensure we don't break any coupling from this gotham conjure generator plugin.

Personally this feels gross and suboptimal to me, so as an alternative here are some classloaders:

## After this PR
==COMMIT_MSG==
gradle-conjure now detects when a generator is actually written in java and allows it to run in a special sandboxed in-process classloader, avoiding the cost of starting up a new JVM.
==COMMIT_MSG==

Real examples.
- log-receiver went from [12s](https://gradle.p.b/s/ndebwv2lzugak/timeline?details=yfeemmyskkaki) -> [3s](https://gradle.p.b/s/lphcnkuzigbwg/timeline?details=yfeemmyskkaki), most of which is the typescript task

## Possible downsides?
- classloaders are one of the more subtle parts of java. Future maintainers may hate me for this
- the start script parsing stuff is pretty fragile, and it would fail if gradle decide to change the layout of their scripts
- securitymanager provides _some_ sandboxing, but doesn't prevent people leaving daemon threads around (I think).
- will this leak resources eventually?


carterkozak (commented):
This is a lot of code in order to invoke the conjure generator library as a library -- I'm not convinced that this is better than providing a simple java api for code-gen in addition to rfc002. Let's chat it over tomorrow :-)

iamdanfox:
Closing out because I thought this was going to be nice and it turned out gross. Ah well, at least I learned a bit more about classloaders :shrug:

iamdanfox:
Annoyingly this is working seamless locally and also from a pTML, but clearly something about the nebula `unit-test` job on CI isn't working :/

iamdanfox:
As discussed in our standup, I think we've hit the end of the timebox here and while it's working fine locally and on java 14, i can't get past that last red test

iamdanfox:
For anyone following this saga, ckozak has rejigged this to not need the security manager stuff here: https://github.com/palantir/gradle-conjure/pull/675

PR #675, opened 2020-12-08
dfox go-fast-hacks modified to use bytebuddy instead of a global securitymanager

carterkozak:
## Before this PR
new JVMs everywhere

## After this PR

==COMMIT_MSG==
in-process java code-generation to reduce jvm startup cost
==COMMIT_MSG==

## Possible downsides?
There's work to do to make this worthwhile in large projects. We're currently rebuilding the classpath for each invocation rather than building it once to allow optimization to help subsequent executions. Still likely better than creating entirely new JVMs, but not ideal.

This implementation assumes that the main class is the only place that can invoke System.exit. This isn't always true, but it's poor form to exit outside of a main method/class and this is probably reasonable enough.

I've not benchmarked this.


gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -47,4 +95,52 @@ static void exec(Project project, String failedTo, List<String> unloggedArgs, Li
...
+                    String.format("Failed to %s. The command '%s' failed.", failedTo, combinedArgs), t);
+        }
+    }
+
+    private static Optional<Method> getMainMethod(URLClassLoader classLoader, String mainClassName) {
+        try {
+            ClassFileLocator locator = new ClassFileLocator.ForUrl(classLoader.getURLs());
+            TypePool typePool = TypePool.ClassLoading.of(classLoader);
+            Class<?> mainClass = new ByteBuddy(ClassFileVersion.ofThisVm(ClassFileVersion.JAVA_V8))
+                    .redefine(typePool.describe(mainClassName).resolve(), locator)
```
iamdanfox, line 126: I had not idea I could solve all my problems by just calling `redefine` this is incredible I love it

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -47,4 +95,52 @@ static void exec(Project project, String failedTo, List<String> unloggedArgs, Li
...
+                    .name(mainClassName + "RedefinedForGradleConjure")
+                    .visit(new ForDeclaredMethods()
+                            .invokable(
+                                    ElementMatchers.any(),
+                                    MemberSubstitution.relaxed()
+                                            .method(ElementMatchers.is(System.class.getMethod("exit", int.class)))
+                                            .replaceWith(GradleExecStubs.getStubMethod())))
```
iamdanfox, line 133: So I'm actually a little surprised that this works, because in conjure-java we don't actually have a System.exit() call in the top level mainclass, and instead rely on picocli and all its reflection... https://github.com/palantir/conjure-java/blob/develop/conjure-java/src/main/java/com/palantir/conjure/java/cli/ConjureJavaCli.java#L52-L54

iamdanfox: how expensive would it be to just visit everything and redefine System.exit everywhere?

carterkozak: The conjure cli uses `System.exit(new CommandLine(new ConjureCli()).execute(args));` while the conjure-java cli uses the deprecated `CommandLine.run(new ConjureJavaCli(), args);`. The latter returns on success and throws on error, it never actually calls System.exit.

> how expensive would it be to just visit everything and redefine System.exit everywhere?

Probably relatively expensive, I think we're safe as is since we recommend against calling System.exit from other classes.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -16,23 +16,71 @@
...
+
+        // We run java things *in-process* to save ~1sec JVM startup time (helpful if there are 100 conjure projects)
+        Optional<ReverseEngineerJavaStartScript.StartScriptInfo> maybeJava =
+                ReverseEngineerJavaStartScript.maybeParseStartScript(executable.toPath());
+        if (maybeJava.isPresent()) {
+            ReverseEngineerJavaStartScript.StartScriptInfo info = maybeJava.get();
+            project.getLogger().info("Running in-process java with args: {}", loggedArgs);
+
+            try (URLClassLoader classLoader =
+                    new URLClassLoader(info.classpathUrls(), GradleExecUtils.class.getClassLoader())) {
```
carterkozak: This should be a child-first loader

iamdanfox:
Ok so I just did a pTML locally and tried this on the biggest foundry repo, and it provides a pretty solid improvement.

`git clean -xdf && ./gradlew compileConjure -i -Dorg.gradle.caching=false` went from ~30 seconds -> ~20 seconds, with a decent amount of the leftover time being spent on typescript and extracting conjure tgzs.

I think I'd be happy to run with this, as I think this codepath is nicely isolated, so we can always back out pretty easily :+1:

carterkozak:
I think we are only getting a fraction of the upside without reusing the classloader, but I'm happy to try it out. Once we have a cached unpack location we should be able to reuse classloaders for greater efficiency.

carterkozak:
:+1:

PR #678, opened 2020-12-10
WIP Attempt using transformers

carterkozak:
==COMMIT_MSG==
Attempt using transformers
==COMMIT_MSG==


gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -27,6 +27,7 @@ private GradleExecUtils() {}
 
     static void exec(Project project, String failedTo, List<String> unloggedArgs, List<String> loggedArgs) {
         List<String> combinedArgs = ImmutableList.<String>builder()
+                .add("sh")
```
carterkozak: breaks windows support, works around our untar not retaining executable bits.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureAttributes.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.conjure;
+
+import org.gradle.api.attributes.Attribute;
+
+final class ConjureAttributes {
+
+    static final Attribute<String> ARTIFACT_TYPE = Attribute.of("artifactType", String.class);
```
carterkozak: delete

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -147,20 +155,36 @@ private static void setupConjureJavaProject(
         if (JAVA_PROJECT_SUFFIXES.stream()
                 .anyMatch(suffix -> project.findProject(project.getName() + suffix) != null)) {
             Configuration conjureJavaConfig = project.getConfigurations().maybeCreate(CONJURE_JAVA);
-            File conjureJavaDir = new File(project.getBuildDir(), CONJURE_JAVA);
+            conjureJavaConfig.getAttributes().attribute(ARTIFACT_TYPE, "tgz");
```
ferozco, line 158: I'm not sure you need to apply the attribute on the config if you're going to use an artifact view

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -214,7 +239,8 @@ private static void setupConjureObjectsProject(
             Supplier<GeneratorOptions> optionsSupplier,
             Task compileConjure,
             TaskProvider<?> compileIrTask,
-            ExtractExecutableTask extractJavaTask) {
+            ArtifactView view,
```
ferozco, line 242: you might want to pass a `Provider<File>` instead of constructing it within the method

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
carterkozak: I think the input artifact type must be tgz for any of this to work, so this should constrain the configuration to only tgz artifacts if I understand correctly. The docs leave a lot to be desired though, and mention that `artifactType` is "special".

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
carterkozak: Do you mean in place of the view, or combination of the view + executable name? Unfortunately later in the pipeline we have some special code to apply a ".bat" suffix on windows which could be refactored elsewhere, but makes the change larger.

ferozco (commented):
overall the approach seems reasonable, we could alternatively pull out the extract executables task to the root project and accomplish the same sharing that way

carterkozak:
> we could alternatively pull out the extract executables task to the root project and accomplish the same sharing that way

I like the single-task approach a little better, but it would be more painful to roll out and likely cause more friction since we don't apply a conjure plugin to the root project.

ferozco:
We can have the per-project plugin apply a plugin on the root. it wouldn't require any change in peoples scripts

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
ferozco: I think we probably want to centralize the logic for windows compat

carterkozak: I agree that the windows handling is done poorly, I'd really prefer if that was fixed separately as I don't like to change two things in the same PR.

I do not know what you mean by a `Provider<File>` or how to build one. Gradle provides several types of provider implementations based on files and directories, but I don't see any way to create one from an ArtifactView.

Extracting the executable doesn't need to be aware of the executable script name, the transformation requires the entire extracted tarball which we could later drill down to the executable. We wouldn't want to pull a cached transformation result that neglects to bring jars along with it. Perhaps I've misinterpreted your comment?

PR #684, opened 2020-12-16
Centralize generator extraction onto the root project

carterkozak:
## Before this PR
Each conjure module would separately extract the conjure compiler and all generators.

## After this PR
==COMMIT_MSG==
Several standard generators are extracted to the project root
==COMMIT_MSG==

## Possible downsides?
`conjureCompiler` and `conjureJava` gradle configurations are not longer created on subprojects, these are only used on the root project. Projects which use these configurations will break unless they move the dependencies to the root project.
This shouldn't impact anyone using GCV where versions are defined globally. It is no longer possible to use different conjure versions across subprojects, but that was never supported using GCV.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -146,11 +146,8 @@ public final class ConjurePlugin implements Plugin<Project> {
             ConjureProductDependenciesExtension productDependencyExt) {
         if (JAVA_PROJECT_SUFFIXES.stream()
                 .anyMatch(suffix -> project.findProject(project.getName() + suffix) != null)) {
-            Configuration conjureJavaConfig = project.getConfigurations().maybeCreate(CONJURE_JAVA);
-            File conjureJavaDir = new File(project.getBuildDir(), CONJURE_JAVA);
-            project.getDependencies().add(CONJURE_JAVA, CONJURE_JAVA_BINARY);
-            ExtractExecutableTask extractJavaTask = ExtractExecutableTask.createExtractTask(
-                    project, "extractConjureJava", conjureJavaConfig, conjureJavaDir, "conjure-java");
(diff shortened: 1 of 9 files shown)


carterkozak:
I didn't migrate over local codegen or python generators yet. Migrating custom generators is more difficult because dependencies are defined in subprojects using the `conjureGenerators` configuration and would require buildscript changes to update.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -146,11 +146,8 @@ private static void setupConjureJavaProject(
             ConjureProductDependenciesExtension productDependencyExt) {
         if (JAVA_PROJECT_SUFFIXES.stream()
                 .anyMatch(suffix -> project.findProject(project.getName() + suffix) != null)) {
-            Configuration conjureJavaConfig = project.getConfigurations().maybeCreate(CONJURE_JAVA);
-            File conjureJavaDir = new File(project.getBuildDir(), CONJURE_JAVA);
-            project.getDependencies().add(CONJURE_JAVA, CONJURE_JAVA_BINARY);
-            ExtractExecutableTask extractJavaTask = ExtractExecutableTask.createExtractTask(
-                    project, "extractConjureJava", conjureJavaConfig, conjureJavaDir, "conjure-java");
+            ExtractExecutableTask extractJavaTask = ExtractExecutableTask.getOrCreateExtractTaskOnRootProject(
```
ferozco: I would create a separate plugin that registers the task and have this snippet can apply the plugin on the root project

carterkozak: Makes sense, does gradle deduplicate plugin registration if it's called several times?

ferozco: yes it handles it for you!

carterkozak:
Should be good to go

ferozco (approved):

PR #685, opened 2020-12-16
Conjure generators run in-process reuse classloaders between runs

carterkozak:
## Before this PR
Code was loaded for each generator invocation.

## After this PR
This reduces overhead of the conjure generator when a single generator
is reused, for example to generate java objects, dialogue clients, and
undertow servers.

This allows optimizations based on generating conjure objects in one project to improve performance in the next.

==COMMIT_MSG==
Conjure generators run in-process reuse classloaders between runs
==COMMIT_MSG==

## Possible downsides?
One of the two hardest problems in CS...


.circleci/config.yml
```diff
@@ -125,8 +125,8 @@ jobs:
     environment:
       CIRCLE_TEST_REPORTS: /home/circleci/junit
       CIRCLE_ARTIFACTS: /home/circleci/artifacts
-      GRADLE_OPTS: -Dorg.gradle.jvmargs='-XX:MaxMetaspaceSize=256m' -Dorg.gradle.workers.max=2
```
iamdanfox: this will get overwritten by the circle template... is there a number we could dial this up to that would make this stuff pass ? alternatively would the forkEvery option just spin up more JVMs?

carterkozak: Not if the circle template can't run tests ;-)
Agree this isn't the long term solution, I just wanted the tests to run without hitting resource limits.

iamdanfox:
So the build scans show this clearly works... just wondering whether we're literally re-implementing this first-class gradle concept:

https://docs.gradle.org/current/userguide/build_services.html

It even says you can make em implement AutoCloseable and it'll just do the right thing??? I'm thinking this _might_ appear really nicely in the build scan UI, and will possibly play nicely with their upcoming configuration cache work

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -40,35 +45,133 @@
...
 
     static void exec(
             Project project, String failedTo, File executable, List<String> unloggedArgs, List<String> loggedArgs) {
+        getRunner(executable).invoke(project, failedTo, unloggedArgs, loggedArgs);
+    }
+
+    private static ConjureRunner getRunner(File executable) {
+        File key = executable.getAbsoluteFile();
+        while (true) {
+            ConjureRunner conjureRunner = runnerCache.get(key);
```
iamdanfox: given that this is a loading cache, if we somehow managed to make the `load` method which _always_ returned a not 'valid' instance, this could loop forever right?

carterkozak: That's correct.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -16,130 +16,30 @@
...
-
-            return Optional.of(mainClass.getMethod("main", String[].class));
-        } catch (ReflectiveOperationException e) {
-            log.warn("Failed too get main method {}", mainClassName, e);
-            return Optional.empty();
-        }
+        project.getGradle()
+                .getSharedServices()
+                .registerIfAbsent(
+                        "conjure-runner-" + executable,
```
iamdanfox, line 35: maybe this is what we were slacking about, but if you didn't have this `executable` concatenated in the name here, do all tasks try to use the same SharedService???

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java
```diff
@@ -0,0 +1,232 @@
...
+            this.mainMethod = mainMethod;
+            this.classLoader = classLoader;
+            this.executableLastModifiedTime = executable.lastModified();
+        }
+
+        @Override
+        public boolean isValid() {
+            // If the file has been modified, this data is no longer valid;
+            return executableLastModifiedTime == executable.lastModified();
+        }
```
iamdanfox, line 178: is this isValid stuff _strictly_ necessary? the executable includes the version number in its path right?

carterkozak: We can probably remove it, I just wanted to be careful. The most likely case is that something has deleted the location where we extracted our dist. Now that we're using shared services it's less likely.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
carterkozak: Yep. I updated this with a comment a few minutes ago. The `registerIfAbsent` method _only_ takes the name into account, not even the service type. Initially my tests failed attempting to run the conjure compiler with args to generate objects...

iamdanfox (approved):
Looks really clean, thank you for diving into the Gradle world to land these speedups - @rzpt I hope we'll be able to leverage these speedups in the 'large internal project'!

(personally I'd probably ship this as a changelog 'break' because version numbers are free)

carterkozak:
Good call, I've updated the changelog type now that we require 6.x

PR #686, opened 2020-12-16
Preserve the `conjureTypeScript` for typescript publishing

carterkozak:
## Before this PR
`compileConjure` fails anywhere that uses the internal ts publishing plugin.

## After this PR
==COMMIT_MSG==
Preserve the `conjureTypeScript` for typescript publishing
==COMMIT_MSG==

## Possible downsides?
cruft

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractConjurePlugin.java
@@ -17,6 +17,7 @@
 package com.palantir.gradle.conjure;
 
 import java.io.File;
+import java.util.Objects;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractConjurePlugin.java
```diff
@@ -65,7 +66,14 @@ static ExtractExecutableTask applyConjureJava(Project project) {
     }
 
     static ExtractExecutableTask applyConjureTypeScript(Project project) {
-        return applyAndGet(project, ExtractConjureTypeScriptPlugin.class, ExtractConjureTypeScriptPlugin.TASK_NAME);
+        ExtractExecutableTask result =
+                applyAndGet(project, ExtractConjureTypeScriptPlugin.class, ExtractConjureTypeScriptPlugin.TASK_NAME);
+        // Preserve the conjureTypescript configuration so publishing works
+        if (!Objects.equals(project, project.getRootProject())) {
```
ferozco, line 72: We'll be resolving the dependency twice but that should basically be free, so this is probably fine. Think we probably want to be smarter in internal-publish though

ferozco (approved):

PR #706, opened 2021-01-20
Feature/flex project structure

esword:
## Before this PR

We have a very large project with multiple conjure apis.  Unfortunately, we have a build tooling that requires a flat project structure, so cannot use the gradle plugin directly. 

## After this PR

Derived projects can be on the same level as the api project

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -135,7 +135,7 @@ public final class ConjurePlugin implements Plugin<Project> {
             TaskProvider<?> compileIrTask,
             ConjureProductDependenciesExtension productDependencyExt) {
         if (JAVA_PROJECT_SUFFIXES.stream()
-                .anyMatch(suffix -> project.findProject(project.getName() + suffix) != null)) {
+                .anyMatch(suffix -> findDerivedProject(project, project.getName() + suffix) != null)) {
             ExtractExecutableTask extractJavaTask = ExtractConjurePlugin.applyConjureJava(project);
 
             setupConjureObjectsProject(project, optionsSupplier, compileConjure, compileIrTask, extractJavaTask);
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -584,6 +587,36 @@ private static void setupGenericConjureProjects(
...
+
+        boolean useFlatProjectStructure = project.hasProperty(USE_FLAT_PROJECT_STRUCTURE_PROPERTY);
+        if (!useFlatProjectStructure) {
+            genericSubProjects = Maps.filterKeys(project.getChildProjects(), childProjectName -> {
+                return childProjectName.startsWith(projectName)
+                        && !FIRST_CLASS_GENERATOR_PROJECT_NAMES.contains(
+                                extractSubprojectLanguage(projectName, childProjectName));
+            });
+        } else {
+            if (project.hasProperty(GENERIC_GENERATOR_LANGUAGE_NAMES_PROPERTY)) {
```
bjlaub, line 606: this is necessary for our use case because with the flat project structure, we can't actually infer the difference between a top-level Conjure API project, where this plugin is applied, and a pseudo-subproject that receives the output of a custom generator.

As an example, we have both a `foo-api` and a `foo-api-internal` project, both of which define Conjure APIs, and both of which use a custom generator (so we also have `foo-api-custom` and `foo-api-internal-custom`). Without the explicit hint here, `foo-api-internal` could be interpreted as a pseudo-subproject of `foo-api`, for the "internal" custom generator type.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -550,8 +553,8 @@ private static void setupGenericConjureProjects(
                 if (!FIRST_CLASS_GENERATOR_PROJECT_NAMES.contains(conjureLanguage)
                         && !generators.containsKey(conjureLanguage)) {
                     throw new RuntimeException(String.format(
-                            "Discovered subproject %s without corresponding " + "generator dependency with name '%s'",
-                            subproject.getPath(), ConjurePlugin.CONJURE_GENERATOR_DEP_PREFIX + subprojectName));
+                            "Discovered subproject %s without corresponding generator dependency with name '%s'",
+                            subproject.getPath(), ConjurePlugin.CONJURE_GENERATOR_DEP_PREFIX + conjureLanguage));
```
carterkozak, line 557: not new in this change, but this is an odd mix of string-format and concatenation.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
carterkozak: Is there any way we can invert this logic to check the `conjureGenerators` dependencies first to compute a list of possible conjure module names? That way we don't have to worry about suffix ambiguity.

bjlaub: My understanding is that we can't do that because the code here: https://github.com/palantir/gradle-conjure/blob/develop/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java#L516-L585 needs to resolve the `conjureGenerators` configuration in order to get a list of names, which has to happen after the projects are evaluated. But, that code also needs to setup tasks on the project that will receive the generator output.

@AlexLandau had a separate idea of separating this code out into a different gradle plugin that is actually applied on the receiver project (instead of the root API project), and placing the `conjureGenerators` configuration there. I think that would solve the problem for our flat project structure use case, with the tradeoff being that it's more boilerplate to have to manually add.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
esword: yes, there is also inconsistency in the structure of the various pre-checks for the different types of projects.  I originally revamped those to be consistent, but wanted the number of changes to be as small as possible.

esword:
Closing in favor of https://github.com/palantir/gradle-conjure/pull/707 - same code, just PR from branch rather than fork

PR #707, opened 2021-01-22
Allow derived projects to be at peer level to api project

esword:
## Before this PR
We have a very large project with multiple conjure apis. Unfortunately, we have a build tooling that requires a flat project structure, so cannot use the gradle plugin directly.

This PR should replace 
https://github.com/palantir/gradle-conjure/pull/706

It's the same code, but not pushing from a fork to fix circle issues

## After this PR
Derived projects can be on the same level as the api project


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -83,6 +85,12 @@
     /** Make the old Java8 @Generated annotation available even when compiling with Java9+. */
     static final String ANNOTATION_API = "jakarta.annotation:jakarta.annotation-api:1.3.5";
 
+    /** Tells plugin to look for derived projects at same level as the api project rather than as child projects. */
+    static final String USE_FLAT_PROJECT_STRUCTURE_PROPERTY = "conjureUseFlatProjectStructure";
```
CRogers: We normally like to prefix our gradle properties with `com.palantir.`, so maybe make this `com.palantir.conjure.use-flat-project-structure`?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -83,6 +85,12 @@
...
+
+    /** Tells plugin the names of generic generator derived projects when in flat mode. */
+    static final String GENERIC_GENERATOR_LANGUAGE_NAMES_PROPERTY = "conjureGenericGeneratorLanguageNames";
```
CRogers: Ditto with the prefix

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -633,4 +666,29 @@ static Task createWriteGitignoreTask(Project project, String taskName, File outp
...
+        }
+        return derivedProject.getPath();
+    }
+
+    private static Project findDerivedProject(Project project, String projectName) {
+        boolean useFlatProjectStructure = project.hasProperty(USE_FLAT_PROJECT_STRUCTURE_PROPERTY);
+        if (!useFlatProjectStructure) {
+            return project.findProject(projectName);
+        } else {
+            if (project.getParent() == null) {
```
CRogers: Could this section be `project.getRootProject().findProject(...)` avoiding the null check?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -584,6 +587,36 @@ private static void setupGenericConjureProjects(
...
+     * Locates projects either as child projects or as peer projects whose names match the patterns given by
+     * the GENERIC_GENERATOR_LANGUAGE_NAMES_PROPERTY property.
+     */
+    private static Map<String, Project> findGenericDerivedProjects(Project project) {
+        Map<String, Project> genericSubProjects = Collections.emptyMap();
+        String projectName = project.getName();
+
+        boolean useFlatProjectStructure = project.hasProperty(USE_FLAT_PROJECT_STRUCTURE_PROPERTY);
+        if (!useFlatProjectStructure) {
+            genericSubProjects = Maps.filterKeys(project.getChildProjects(), childProjectName -> {
```
CRogers: style nit: Can we return early to avoid having to define the variable? Can also get rid of the else to lower the nesting a bit.

changelog/@unreleased/pr-707.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: "Allow derived projects to be at peer level to api project\n\n## Before this PR\nWe have a very large project with multiple conjure apis. Unfortunately, we have a build tooling that requires a flat project structure, so cannot use the gradle plugin directly.\n\nThis PR should replace \nhttps://github.com/palantir/gradle-conjure/pull/706\n\nIt's the same code, but not pushing from a fork to fix circle issues\n\n## After this PR\nDerived projects can be on the same level as the api project\n\n## Possible downsides?\n<!-- Please describe any way users could be negatively affected by this PR. -->"
```
CRogers: Probably need a bit of a shorter changelog, just edit the commentbot comment and recheck the box.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
esword: heh - that is definitely a style difference for me.  I prefer single point of return when there is a possibility of a hanging "else".  But when in Rome...

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
esword: we could, but I didn't really want for it to go all the way to root in the circumstance where api is possibly embedded in something else.  But that doesn't impact my use case, so happy to change it.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
esword: I did it, but just so you are aware, it forces this type of syntax when adding the property in a build file - project.ext['com.palantir.conjure.use_flat_project_structure']=true

AlexLandau:
I expect this won't work with configuration-on-demand unless you add `evaluationDependsOn(':my-conjure-api')` to all the generator projects.

CRogers:
@AlexLandau Ah, just to be clear, before since they were guaranteed to be evaluated first because they were subprojects, now they need an explicit `evaluationDependsOn` to work?

This should only break configuration on demand if you are using this new feature flag right? If so, I'll leave it up to you to decide if it's ok to ship this without c-o-d support, considering this is only going to be used in big-internal-project.

esword:
Alex is right that it does require an extra config to work with c-o-d, but I still prefer to move to the standard plugin.  We'll have additional config we have to feed it anyway - we use a fair number of the command line options.  Think it's good to go @CRogers ?

CRogers (approved):

CRogers:
Hmmm, I don't know why the circle build statuses are not being reported

AlexLandau:
~Does this actually work for the intended repo, given that the `-typescript` projects are in a different part of the directory structure?~

EDIT: Never mind, I see that it's using project paths rather than directory paths.

CRogers:
Closing + reopening to see if that fixes the circle issue

CRogers:
Nope, evidently not

esword:
Closing because of circle issues.  Changed merged with https://github.com/palantir/gradle-conjure/pull/710

PR #710, opened 2021-01-26
Handle derived projects being at same level as conjure api project

esword:
## Before this PR
Redoing PR 707 - https://github.com/palantir/gradle-conjure/pull/707

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


fawind (approved):
For reference: Previous discussion on https://github.com/palantir/gradle-conjure/pull/707

PR #759, opened 2021-04-26
Allow extensions to be set on CompileIrTask

esword:
## Before this PR
Projects could not add extension information to IR files

## After this PR

==COMMIT_MSG==
Allow extensions to be set on CompileIrTask
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
@@ -34,6 +34,7 @@ import org.gradle.api.provider.SetProperty;
 import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputDirectory;
+import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.OutputFile;
@@ -52,6 +53,7 @@ public class CompileIrTask extends DefaultTask {
             getProject().getObjects().setProperty(ServiceDependency.class);
     private final MapProperty<String, Serializable> conjureExtensions =
             getProject().getObjects().mapProperty(String.class, Serializable.class);
+    private final RegularFileProperty extensionsFile = getProject().getObjects().fileProperty();
 
     public CompileIrTask() {
         conjureExtensions.convention(new HashMap<>());
@@ -114,6 +116,13 @@ public class CompileIrTask extends DefaultTask {
         return conjureExtensions;
     }
 
+    @InputFile
+    @Optional
+    @PathSensitive(PathSensitivity.RELATIVE)
+    public final RegularFileProperty getExtensionsFile() {
+        return extensionsFile;
+    }
+
     @TaskAction
     public final void generate() {
         File executable = new File(executableDir.get(), EXECUTABLE);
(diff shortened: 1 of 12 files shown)


esword:
I am going to add an option to load extensions from a json file in map format.  makes it easier to cache and chain with other tasks,

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -97,6 +109,19 @@ public final File getExecutableDir() {
...
+    @Input
+    @Optional
+    public final MapProperty<String, Serializable> getConjureExtensions() {
+        return conjureExtensions;
+    }
+
+    @InputFile
+    @Optional
+    @PathSensitive(PathSensitivity.RELATIVE)
+    public final RegularFileProperty getExtensionsFile() {
```
fawind, line 121: Do we need both the input file and property for now? Might be worth just using one for now to keep the API minimal (maybe the file input?). But no strong opinion here.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -97,6 +109,19 @@ public final File getExecutableDir() {
         return productDependencies;
     }
 
+    @Input
+    @Optional
+    public final MapProperty<String, Serializable> getConjureExtensions() {
```
fawind, line 114: @carterkozak flagged that this could break when the task is invoked from command line and e.g. the command line has a max limit for input args if the extension data exceeds the limit.

But if it's working for you project we can handle such issues when they arise -- just something to keep in mind.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
esword: I think I'm okay with just the file for production, but the inline feels more complete and I have a hunch it may make some testing easier.  Haven't gotten all the way through testing with local version yet to know for sure.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
esword: You mean when passing to the actual generator?  I just wrote a quick test that passes an extension of ~9M and it worked on my local mac, which I presume means it will work on linux.  There might be an issue on Windows, but I think that would be a rare case and, as you say, can address if it comes up.

fawind: Thanks for adding the test! Yes, when passing it to the generator executable here:

https://github.com/palantir/gradle-conjure/blob/cbaa18cd6056459b365d369d6cb2d0d49067ad8a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java#L132-L133

Yep, might only be an issue in specific environments (windows maybe?). Should be good for now, we can think about this more if it becomes an actual problem (e.g. pass the extension by file to the executable).

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
fawind: Sounds ok when it makes testing easier.

fawind (approved):

PR #761, opened 2021-04-29
remove duplicate code for java project init

esword:
## Before this PR
the 5 java project types had about 98% duplicated code to init them.  I couldn't take it any more.

## After this PR

==COMMIT_MSG==
Unified init code, with consumer passed in for the couple lines of customization for each.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -18,8 +18,10 @@ package com.palantir.gradle.conjure;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.palantir.gradle.conjure.api.ConjureExtension;
 import com.palantir.gradle.conjure.api.ConjureProductDependenciesExtension;
 import com.palantir.gradle.conjure.api.GeneratorOptions;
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -136,246 +138,134 @@ public void apply(Project project) {
...
-            Task cleanTask = project.getTasks().findByName(TASK_CLEAN);
-            cleanTask.dependsOn(project.getTasks().findByName("cleanCompileConjureRetrofit"));
-            subproj.getDependencies().add("api", findDerivedProject(project, objectsProjectName));
-            subproj.getDependencies().add("api", "com.google.guava:guava");
-            subproj.getDependencies().add("api", "com.squareup.retrofit2:retrofit");
-            subproj.getDependencies().add("compileOnly", ANNOTATION_API);
+            Task cleanTask = parentProject.getTasks().findByName(TASK_CLEAN);
+            cleanTask.dependsOn(parentProject.getTasks().findByName("cleanCompileConjure" + upperSuffix));
+            if (isNotObjectsProject) {
+                ConjureJavaServiceDependencies.configureJavaServiceDependencies(subproj, productDependencyExt);
```
carterkozak, line 231: I think this should only apply to client interfaces: dialogue, jersey, and retrofit. I'm not sure why we ever added it to conjure-undertow, probably bad copy-pasta from the jersey generator.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -136,246 +138,134 @@ public void apply(Project project) {
...
+                subproj.getDependencies().add("api", findDerivedProject(parentProject, objectsProjectName));
```
carterkozak, line 232: Can we remove the `findDerivedProject` flat-project-structure code now that our large internal project has added nested projects?

carterkozak (approved):
Change looks good to me, I defer to your judgement if you'd like to handle comments in this PR or separately.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
esword: probably, but I want to wait a little bit more to make sure we don't have a catastrophic problem.  will do a separate one.

Issue #760, opened 2021-04-28
Copy IR into target projects as resource


esword:
## What happened?
I am working on something that does build-time analysis in java projects that consume conjure APIs to determine what endpoints are called.  I need access to the IR for this to work.  I have a hacky way to figure out the name of the IR artifact for a given conjure-jar dependency and go download it (e.g. if a project has 'my-company:my-api-dialogue' as a dependency, I can figure out that the IR artifact is 'my-company:my-api' and search repostories for where to download it).  It would be much cleaner though if the IR was in the jar itself as a resource file.

## What did you want to happen?
Put the conjure.json file in the jar at the root level, preferably with a standard name like just "conjure.json".  Or, if there will ever be multiple conjure.json files in a jar, could just keep them with the name of the generated file directly as it comes from the compileIr task.

## Proposed Solutions
I have experimented with a couple of ways to do this:

1. Add the output directory of the compileIR task directly as a resource directory of the derived projects.  Can also make tweaks to the ConjureJavaLocalCodegenPlugin and ConjureLocalPlugin to pass in their source IRs as well in a similar way.  This takes the least amount of code, but is not canonical gradle because the derived project is then reaching outside of it's project directory for some of it's resources.

2. Add a new copyIrToXXX task to the derived java projects (objects, retrofit, jersey, etc) that copies the IR into src/generated/resources.  This is the most obvious way to do things and doesn't require much code, but it is more than the first option and also doesn't easily extend to generic generators (which I have)

3. Add an optional to property to ConjureGeneratorTask for irOutputDirectory.  If set, the generator task will copy over the source IRs to the directory.  This is the least intrusive and makes it easy for anything else that configures a generator (including the local stuff and generic generators) to copy the IR.

Thoughts?

carterkozak:
Do we need the entire IR with object definitions, or just the mapping from data that uniquely identifies an endpoint (`GET /foo`)  to the minimum product dependency version for that endpoint? Could we add the minimal mapping the same way we handle overall product dependency info?
https://github.com/palantir/gradle-conjure/blob/26d43e19fba43f19868cc425136cade8e1b12656/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java#L35-L42
[...]

PR #762, opened 2021-04-29
support for writing endpoint minimum version info to jar

esword:
## Before this PR
this is a solution to the specific case needed for issue #760 

## After this PR

==COMMIT_MSG==
support for writing endpoint minimum version info to jar
==COMMIT_MSG==


gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointMinimumVersion.java
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.conjure.api;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+import java.util.Objects;
+
+public final class EndpointMinimumVersion implements Serializable {
```
carterkozak, line 23: We tend to use immutable objects in most places rather than mutable bean style objects -- I think this project already use the immutables annotation processor, which may make it easier.

esword: I did that at first, but this object is (possibly) read in using gradle closure processing and I couldn't get that to work with Immutables because it uses direct reflection to set fields, not invoke the methods.  It technically works if you pass it a builder, but then the builder fails because the bitfield to indicate fields are set is not cleared.  I presume that's why the ServiceDependency object is the same.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConfigureEndpointMinimumVersionsTask.java
```diff
@@ -0,0 +1,81 @@
...
+import org.gradle.api.java.archives.Manifest;
+import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.api.tasks.bundling.Jar;
+
+public class ConfigureEndpointMinimumVersionsTask extends DefaultTask {
+    private final SetProperty<EndpointMinimumVersion> endpointVersions =
```
carterkozak: Should we use a `List` here? I don't think we need to hash each value, not do we expect duplicates with the same verb/path and different versions.
Very supportive of the collection here rather than a `Map`, since it gives us more flexibility in the future.

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointMinimumVersion.java
carterkozak: You're right, that makes sense

carterkozak:
Should the collection of `EndpointMinimumVersion` values be encapsulated within the `ServiceDependency`?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConfigureEndpointMinimumVersionsTask.java
esword: done

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -575,4 +584,16 @@ private static Project findDerivedProject(Project project, String projectName) {
...
+    private static void configureEndpointMinimumVersions(
+            Project project, EndpointMinimumVersionsExtension minimumVersionsExtension) {
+        project.getPluginManager().apply(EndpointMinimumVersionsPlugin.class);
+        project.getTasks()
+                .named(
+                        EndpointMinimumVersionsPlugin.CONFIGURE_ENDPOINT_MINIMUM_VERSIONS_TASK,
+                        ConfigureEndpointMinimumVersionsTask.class,
+                        task -> {
+                            task.getVersions().set(minimumVersionsExtension.getEndpointVersions());
+                        });
```
fawind: It's a bit confusing to me that we create the `EndpointMinimumVersionsExtension` in this plugin as well as in the `EndpointMinimumVersionsPlugin` plugin. Is there a reason for this?

Likewise we configure the `ConfigureEndpointMinimumVersionsTask` in this method and in the `EndpointMinimumVersionsPlugin`. Maybe we can get rid of one code path?

esword: yaaa... I have no idea why I did that.  I think too many looks at the recommended product dependencies configuration which is spread out in a few places.  Just applying the plugin should be enough.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureServiceDependencyTest.groovy
```diff
@@ -256,11 +258,73 @@ class ConjureServiceDependencyTest extends IntegrationSpec {
...
+        serviceDependencies {
+            serviceDependency {
+                productGroup = "com.palantir.conjure"
+                productName = "conjure"
+                minimumVersion = "1.2.0"
+                recommendedVersion = "1.2.0"
+                maximumVersion = "2.x.x"
+            }
+        }
+        endpointVersions {
```
fawind: Similar to what Carter mentioned [here](https://github.com/palantir/gradle-conjure/pull/762#issuecomment-830177981), it feels a bit random to have this as its own top-level extension.

Should we move this inside the `serviceDependency` extension ([ref](https://github.com/palantir/gradle-conjure/blob/9752440246345e85add20605512ebb654bdd5abb/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureProductDependenciesExtension.java#L24))? E.g. something like:

```gradle
serviceDependencies {
    serviceDependency {
        // ...
    }
    endpointVersion {
       // ...
    }
}
```

esword: good idea. done

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java
```diff
@@ -16,8 +16,6 @@
 
 package com.palantir.gradle.conjure;
 
-import com.google.common.collect.Iterables;
```
fawind, line 19: I think this import got swallowed during a merge: [circle](https://app.circleci.com/pipelines/github/palantir/gradle-conjure/1234/workflows/7a881f66-5177-442a-8f39-3f6f33ea31d9/jobs/9610)

fawind (approved):

PR #769, opened 2021-05-06
Add support for optional service dependencies

ejarleberg:
## Before this PR
It's currently not possible to specify optional service dependencies. This makes it impossible to have an API
project render multiple optional product dependencies e.g.

## After this PR
An API project can now specify optional service dependencies.
This is useful in the scenario where an API might have several different
implementing services e.g.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java
@@ -16,11 +16,13 @@
 
 package com.palantir.gradle.conjure;
 
+import com.google.common.collect.Iterables;
 import com.palantir.gradle.conjure.api.ConjureProductDependenciesExtension;
 import com.palantir.gradle.conjure.api.ServiceDependency;
 import com.palantir.gradle.dist.ConfigureProductDependenciesTask;
 import com.palantir.gradle.dist.ProductDependency;
 import com.palantir.gradle.dist.RecommendedProductDependenciesPlugin;
+import com.palantir.logsafe.Preconditions;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.gradle.api.Project;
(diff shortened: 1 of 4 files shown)


ferozco:
I believe when optional dependencies were introduced we decided that only the service should be able to make a dependency optional since the service knows whether it can operate with or without the dependency being present and can therefore make the decision while the API producer (i.e. the conjure project) can not

ejarleberg:
Me and @ferozco caught up on Slack. We decided that it actually does make sense for API project to declare optional service dependencies. 
A typical use-case for optional dependencies would be an API that has two different backing stores, from the sls-spec regarding optional dependencies:

> As an example, consider a service "nice-service" that is able to store its underlying data in either Postgres or Alta, but only one of them. Such service should declare optional dependencies on both Postgres and Alta to ensure that the absence of either of them doesn't block its management by the deployability tooling and to ensure that the service actually being used for data storage is in a compatible version.

As a consumer of the API, I do not necessarily care which actual implementation I'm wired up to, as long as the service conforms to the API. In fact, as a consumer, I might not even know of all possible implementations.
The consuming service can easily discover any implementing service of the API (by using a shared discovery produce role across two different implementations of the API). This is exactly how Gotham's Dispatch discovers either Horizon, or Horizon-Live, and it doesn't care which one it consumes, since both implement the same API.

In line with the recommendation from https://github.com/palantir/sls-packaging/blob/develop/readme.md#product-dependencies

> Most of your product dependencies should be inferred automatically from on the libraries you depend on. Any one of these jars may contain an embedded 'recommended product dependency' in its MANIFEST.MF (embedded using the Recommended Product Dependencies Plugin).
> 
> However, you can also use the productDependency block to specify these manually (although this is no longer considered a best-practise).

we do not want to explicitly declare both Horizon and Horizon-Live as optional dependencies in Dispatch, but instead have them be inferred from our dependency on the Horizon API.

Me and @ferozco agreed that we couldn't see a use-case for a single, optional service dependency however, so I went ahead and added a validation now that will throw if this is encountered (as it should indicate a misconfiguration/a misunderstanding).

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java
```diff
@@ -42,13 +44,18 @@ static void configureJavaServiceDependencies(
     }
 
     private static Set<ProductDependency> convertDependencies(Set<ServiceDependency> serviceDependencies) {
+        Preconditions.checkArgument(
+                serviceDependencies.size() != 1
+                        || !Iterables.getOnlyElement(serviceDependencies).getOptional(),
+                "a single optional service dependency is not supported");
```
fawind, line 53: Can we add a comment here with a short summery of the reason you outlined in the PR comment? Otherwise in the future, it might be hard to rediscover why we don't allow a single optional service dep.

Looks good to me otherwise, thanks for outlining the use-case!

ejarleberg: Done, thanks!

ferozco (approved):

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaServiceDependencies.java
fawind: Thanks for adding!

PR #771, opened 2021-05-12
Move ConjureProductDependenciesExtension to api package

ntrinquier:
## Before this PR
`ConjureProductDependenciesExtension` was removed from the `.api` package in 87ac7c954bd28d0231b840090cf43723d77d2364, breaking consumers that relied on it.

## After this PR

==COMMIT_MSG==
Move ConjureProductDependenciesExtension to api package
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-771.v2.yml
+++ b/changelog/@unreleased/pr-771.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Move ConjureProductDependenciesExtension to api package
+  links:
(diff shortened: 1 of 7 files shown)


ntrinquier:
Argh, looks like dependencies are not the same between the two gradle projects...

fawind:
Just saw this, thanks for putting up the PR!

I also opened up a similar PR which fixes the dependencies here: https://github.com/palantir/gradle-conjure/pull/772

Closing this for now!

PR #779, opened 2021-05-19
do not add recommended dependencies to server-side only projects

esword:
## After this PR

==COMMIT_MSG==
do not add recommended dependencies to server-side only projects
==COMMIT_MSG==


carterkozak (approved):
Thanks!

PR #792, opened 2021-06-09
Set output directory for compileTypeScript task

pkoenig10:
## Before this PR
`compileTypeScript` is always run because it does not declare outputs.

Screen Shot 2021-06-09 at 2 52 09 PM

## After this PR
`compileTypeScript` declares it's outputs and will not be re-run if it is up-to-date.

I've verified this locally. Here the build scans:
First build: https://g.p.b/s/3jidjf6nifcea/timeline
Second build: https://g.p.b/s/bilbi3h7eakpa/timeline

### DIFF
--- a/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
+++ b/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
@@ -119,14 +119,14 @@ class ConjurePublishTypeScriptTest extends IntegrationSpec {
         file('api/api-typescript/src/index.js').text.contains('export * from "./api";')
     }
 
-    def 'compileConjureTypeScript is up-to-date when run for the second time'() {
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
```diff
@@ -119,14 +119,14 @@ class ConjurePublishTypeScriptTest extends IntegrationSpec {
         file('api/api-typescript/src/index.js').text.contains('export * from "./api";')
     }
 
-    def 'compileConjureTypeScript is up-to-date when run for the second time'() {
+    def 'compileTypeScript is up-to-date when run for the second time'() {
```
pkoenig10, line 122: `compileConjureTypeScript` is already tested in `ConjurePluginTest`. Given the structure of these tests, I assume this was originally intended to test `compileTypeScript`.

jkozlowski (approved):

PR #793, opened 2021-06-11
Revert optimization that sometimes caused infinite loops in TreeMaps

AlexLandau:
The BuildServicesRegistry is apparently not thread-safe, so we
need to either move this registration to during Gradle configuration
or use some other caching mechanism (e.g. put a cache object on an
extension of the root project).

I have left ConjureRunnerResource intact with the expectation
that it will be either used or removed in a subsequent fix of the
optimization. This PR is just to quickly get to a state where we
don't run into this problem.

==COMMIT_MSG==
Revert a performance optimization for large builds that sometimes resulted in hangs due to infinite loops in TreeMap#get.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
@@ -25,8 +25,8 @@ final class GradleExecUtils {
 
     static void exec(
             Project project, String failedTo, File executable, List<String> unloggedArgs, List<String> loggedArgs) {
-        try {
-            ConjureRunnerResource.createNewRunner(executable).invoke(project, failedTo, unloggedArgs, loggedArgs);
+        try (ConjureRunnerResource.ConjureRunner runner = ConjureRunnerResource.createNewRunner(executable)) {
+            runner.invoke(project, failedTo, unloggedArgs, loggedArgs);
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
@@ -16,32 +16,20 @@
 
 package com.palantir.gradle.conjure;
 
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExecUtils.java
```diff
@@ -16,32 +16,20 @@
...
-                        new Action<BuildServiceSpec<Params>>() {
-                            @Override
-                            public void execute(BuildServiceSpec<Params> spec) {
-                                spec.getParameters().getExecutable().set(executable);
-                            }
-                        })
-                .get()
-                .invoke(project, failedTo, unloggedArgs, loggedArgs);
+        try {
+            ConjureRunnerResource.createNewRunner(executable).invoke(project, failedTo, unloggedArgs, loggedArgs);
```
carterkozak: I think we would need to close the classloader after execution in this case:
```suggestion
        try (ConjureRunner runner = ConjureRunnerResource.createNewRunner(executable)) {
            runner.invoke(project, failedTo, unloggedArgs, loggedArgs);
```

AlexLandau: Good catch, done

carterkozak:
@AlexLandau Do we have an upstream gradle ticket tracking this? I can't imagine the BuildServicesRegistry would be intentionally non-threadsafe.

Happy to cut a release with this for now given the failures we're seeing internally, that behavior is clearly broken. :+1:

AlexLandau:
> 
> 
> @AlexLandau Do we have an upstream gradle ticket tracking this? I can't imagine the BuildServicesRegistry would be intentionally non-threadsafe.

I filed https://github.com/gradle/gradle/issues/17434. It may be intended that registering a shared service only happens during (single-threaded) configuration, with tasks getting access to the service through a Property; that's how you'd have to use it to take advantage of [configuration caching](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution) or [limiting concurrent usage of the service](https://docs.gradle.org/current/userguide/build_services.html#concurrent_access_to_the_service).

PR #818, opened 2021-08-10
Disallow non-generated dirs in main sourcesets

atrocities:
## Before this PR
The conjure-consumer dependency scanning strategy relies on annotations
in Generated Java code to determine minimum service dependencies. This does
not work for any non-generated code.

To prevent unexpected problems, all non-generated code must be excluded
from the Conjure generated Javaprojects.

## After this PR
Gradle will error out whenever it finds a non-generated source directory in any of the Java (e.g. objects, dialogue) generated subprojects

## Possible downsides?
Users must move non generated code outside of the generated subprojects if they want to continue using gradle-conjure

### DIFF
--- a/changelog/@unreleased/pr-818.v2.yml
+++ b/changelog/@unreleased/pr-818.v2.yml
@@ -1,0 +1,20 @@
+type: improvement
+improvement:
+  description: |-
+    Disallow non-generated dirs in main sourcesets
+
+    ## Before this PR
+    The conjure-consumer dependency scanning strategy relies on annotations
+    in Generated Java code to determine minimum service dependencies. This does
+    not work for any non-generated code.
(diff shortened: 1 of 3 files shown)


ferozco (changes requested):
I don't think that we can make this change without breaking a variety of existing usages. Some internal examples are:
- g.p.b/f/assets2
- g.p.b/f/hyperion

More generally it seems like the problem you're describing would also arise if there were any external imports in the conjure definition. is that the case?

Is it possible to make the endpoint versioning more resilient or to fail gracefully if it encounters external types?

atrocities:
Thanks for pointing that out Felipe. Let me do some more discussion on our end to figure what the best next steps are here.

PR #826, opened 2021-08-16
use the task-config avoidance APIs

esword:
## Before this PR
Nearly all tasks in the plugin are  created eagerly

## After this PR

==COMMIT_MSG==
use the task-config avoidance APIs
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
@@ -111,15 +111,14 @@ public final class ConjureLocalPlugin implements Plugin<Project> {
                     task.setOutputDirectory(subproj.file(ConjurePlugin.JAVA_GENERATED_SOURCE_DIRNAME));
 
                     generateConjure.dependsOn(task);
-                    subproj.getTasks().getByName("compileJava").dependsOn(task);
 
                     task.dependsOn(gitignoreConjureJava);
                     task.dependsOn(extractJavaTask);
 
-                    Task cleanTask = project.getTasks().findByName(ConjurePlugin.TASK_CLEAN);
-                    cleanTask.dependsOn(project.getTasks().findByName("cleanGenerateJava"));
                     subproj.getDependencies().add("api", subproj);
                 });
+        subproj.getTasks().named("compileJava").configure(t -> t.dependsOn(generateJava));
+        ConjurePlugin.registerClean(project, "cleanGenerateJava");
         ConjurePlugin.applyDependencyForIdeTasks(subproj, generateJava);
     }
 
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
@@ -34,6 +34,7 @@ import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.plugins.BasePlugin;
 import org.gradle.api.plugins.JavaLibraryPlugin;
+import org.gradle.api.tasks.TaskProvider;
 import org.gradle.util.GUtil;
 
 public final class ConjureLocalPlugin implements Plugin<Project> {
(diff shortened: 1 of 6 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -330,70 +344,76 @@ private static void setupConjureTypescriptProject(
...
+                        .register("publishTypeScript", Exec.class, task -> {
+                            task.setDescription("Runs `npm publish` to publish a TypeScript package "
+                                    + "generated from your Conjure definitions.");
+                            task.setGroup(TASK_GROUP);
+                            task.commandLine(npmCommand, "publish");
+                            task.workingDir(srcDirectory);
+                            task.dependsOn(compileConjureTypeScript);
+                            task.dependsOn(compileTypeScript);
+                        });
+                subproj.getPluginManager().withPlugin("maven-publish", _packagingPlugin -> {
```
fawind: This plugin guard-clause is new, right? Makes sense though.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -521,11 +542,14 @@ static void addGeneratedToMainSourceSet(Project subproj) {
...
-    static void applyDependencyForIdeTasks(Project project, Task compileConjure) {
+    static void applyDependencyForIdeTasks(Project project, TaskProvider<?> compileConjure) {
         project.getPlugins().withType(IdeaPlugin.class, plugin -> {
-            Task task = project.getTasks().findByName("ideaModule");
-            if (task != null) {
-                task.dependsOn(compileConjure);
+            // root project does not have the ideaModule task.  There is unfortunately no
+            // safe way to check for existence with the task avoidance APIs
+            try {
+                project.getTasks().named("ideaModule").configure(t -> t.dependsOn(compileConjure));
```
fawind, line 574: From the [docs](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskCollection.html#named-java.lang.String-):
> Locates a task by name, without triggering its creation or configuration, failing if there is no such object.

Could this end up in a race condition when the idea plugin sets up the task only after gradle-conjure is configured?

fawind: Does this give us an advantage over `findByName`? I.e. `findByName` will also not create the task if it does not exist yet.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
```diff
@@ -88,37 +89,37 @@ private void setupConjureJava(
...
+                                "Unable to generate Java bindings since unsafe options were provided",
+                                generatorOptions.getProperties());
+
+                        return generatorOptions;
+                    });
+                    task.setSource(conjureIrConfiguration);
+                    task.setExecutablePath(extractJavaTask::getExecutable);
+                    task.setOutputDirectory(subproj.file(ConjurePlugin.JAVA_GENERATED_SOURCE_DIRNAME));
+
+                    generateConjure.dependsOn(task);
```
ferozco: it isn't safe to add dependencies on a task within its configuration block if you are registering it lazily since the config lambda is only ever executed if the task is already on the dependency graph. You'll need to pull this out

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
ferozco: I think that this is a bit of a behaviour change. I'm not sure if we actually apply the `maven-publish` plugin on conjure projects

esword: that was a fear of mine. Are there other known sources of the publish task? Or do I have to restructure to check if the task is present? I do that in another place with ideaModule, so can do it here as well if necessary.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
esword: but it will instantiate/initialize the task if it does exist, even if it's only been "registered" and not created. It's an annoying problem that there is no "is a TaskProvider with this name registered" call.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalPlugin.java
esword: nice catch!

esword:
one of the last commits made the "clean" tasks come back as eagerly configured.  debugging.

esword:
Oye - the clean tasks are created immediately not  by us, but by the built in gradle CleanRule, so even though I'm looking for them using the tasks.named(xxx) method, underneath the hood they are eagerly created.  So this PR is correct enough for now since it still delays creation of most tasks.  Will puzzle more to see if can somehow get rid of the others.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePublishPlugin.java
```diff
@@ -31,21 +33,25 @@ public void apply(Project project) {
         project.getPluginManager().apply(MavenPublishPlugin.class);
         project.getPluginManager().apply(ConjurePlugin.class);
 
-        CompileIrTask compileIr = (CompileIrTask) project.getTasks().findByName(ConjurePlugin.CONJURE_IR);
-        if (compileIr == null) {
-            throw new GradleException("Unable to find compileIr task");
+        TaskProvider<CompileIrTask> compileIr = null;
+        try {
+            compileIr = project.getTasks().named(ConjurePlugin.CONJURE_IR, CompileIrTask.class);
+        } catch (UnknownDomainObjectException e) {
```
fawind: Nit: Would move this block into its own method so we don't need to do the `null` initialization and reassignment to a final variable.

esword: done!

fawind (approved):

PR #828, opened 2021-08-17
move conjure generator tasks to task-avoidance APIs

esword:
## After this PR

==COMMIT_MSG==
move conjure generator tasks to task-avoidance APIs
==COMMIT_MSG==

## Possible downsides?
Removing the various "set" methods on tasks could break existing projects that customize tasks. We can either role the major API or (better) mark them as deprecated instead of deleting them. I just deleted initially to see how clean I could get things.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
@@ -24,6 +24,7 @@ import org.gradle.api.Action;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.file.ConfigurableFileTree;
+import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Input;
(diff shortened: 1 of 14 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -309,91 +329,119 @@ private static void ignoreFromCheckUnusedDependencies(Project proj) {
...
+                            task.workingDir(srcDirectory);
+                            task.dependsOn(compileConjureTypeScript);
+                            task.dependsOn(compileTypeScript);
+                        });
+                linkPublish(subproj, publishTypeScript);
             });
         }
     }
 
+    private static void linkPublish(Project project, TaskProvider<?> depTask) {
```
fawind: I'm not sure I fully understand the benefit of this method over `maybeCreate`. In the end we still create a publish task if it doesn't exist yet so the number of created tasks shouldn't change?

esword: one is a TaskProvider though, so config happens later. This doesn't matter for that individual task all that much, but it is a cascade effect. If that task is instantiated, 7 other tasks are also eagerly configured in each project, including the conjure compile tasks. It adds up.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
```diff
@@ -32,16 +33,12 @@
...
-    private final Property<String> packageVersion = getProject().getObjects().property(String.class);
-
-    private File productDependencyFile;
-
+public abstract class CompileConjureTypeScriptTask extends ConjureGeneratorTask {
     public CompileConjureTypeScriptTask() {
         Project project = getProject();
-        packageName.set(getProject().provider(project::getName));
-        packageVersion.set(getProject().provider(() -> project.getVersion().toString()));
+        getPackageName().convention(getProject().provider(project::getName));
```
CRogers, line 39: I'm wondering what the difference between `property.convention` and `property.value/set` do.  Both seem to set default values, but I haven't actually seen `convention` used until just now. I thought they might be name overloads but they do seem to have different behaviour in the code.

CRogers: Ah, from https://docs.gradle.org/7.2/userguide/lazy_configuration.html#applying_conventions:

> Once a value is set, the convention is ignored

So this means if you set a convention after there is a value, it's a no-op, where as `value`/`set` would actually replace the value. Sounds like a useful tool - in this case sense we're calling `convention` right after making the properties it should be equivalent to `value`/`set`.

esword: yes, convention is setting a default value so you can tell if it's explicitly overridden with a set call.  I have always pictured that it would be nice to develop a way where you can collect metrics on if a property is every overridden.  That way you can tell if changing/removing the property would cause a break to consumers.  Just never did anything with the idea.

CRogers (approved):
Thanks for the improvements!

netcazean (commented):
##828

netcazean (commented):
#846

PR #883, opened 2021-11-05
Make max version nullable by adding JsonInclude NON_NULL.

ylee088:
## Before this PR

Previously null/empty max versions defaulted to x.x.x which we don't want.

## After this PR

==COMMIT_MSG==
Allow empty max versions without default values.
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointVersionBound.java
+++ b/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointVersionBound.java
@@ -16,9 +16,12 @@
 
 package com.palantir.gradle.conjure.api;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.io.Serializable;
 import java.util.Objects;
+import java.util.Optional;
 
 public final class EndpointVersionBound implements Serializable {
 
@@ -31,6 +34,7 @@ public final class EndpointVersionBound implements Serializable {
     @JsonProperty("min-version")
     private String minVersion;
 
+    @JsonInclude(Include.NON_ABSENT)
     @JsonProperty("max-version")
     private String maxVersion;
 
@@ -58,12 +62,8 @@ public final class EndpointVersionBound implements Serializable {
         this.minVersion = minVersion;
     }
 
-    public String getMaxVersion() {
-        if (maxVersion != null) {
-            return maxVersion;
-        } else {
-            return "x.x.x";
-        }
+    public Optional<String> getMaxVersion() {
+        return Optional.ofNullable(maxVersion);
     }
 
     public void setMaxVersion(String maxVersion) {
--- a/changelog/@unreleased/pr883.v2.yml
+++ b/changelog/@unreleased/pr883.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Allow empty max versions without default values.
+  links:
+    - https://github.com/palantir/gradle-conjure/pull/883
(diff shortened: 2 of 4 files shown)


gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointVersionBound.java
```diff
@@ -58,12 +60,9 @@ public void setMinVersion(String minVersion) {
         this.minVersion = minVersion;
     }
 
+    @JsonInclude(Include.NON_NULL)
```
carterkozak: Can we add round-trip serialization of this class with and without maxVersion set, for both JSON and java serialization?

Does the `JsonInclude` annotation belong with the `JsonProperty` annotation above on line 36?

ylee088: Sorry, had to step away for a bit after pushing the PR. Will add some tests.

ylee088: Turns out I did add a test (at least one direction of the round-trip, and may need some updates): https://github.com/palantir/gradle-conjure/blob/a387dcb6d0169833c5c15f1d5403e5d3bac2c156/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureServiceDependencyTest.groovy#L383
 
But are the tests actually running? I'm only seeing java test results: https://app.circleci.com/pipelines/github/palantir/gradle-conjure/1608/workflows/797c8a2a-591d-4e8e-a4e7-338ac8df6a07/jobs/11399/artifacts

ylee088: I didn't see any other tests being run locally either but I was able to run all the tests with some gradle hackery I googled. I haven't added the gradle fix because I didn't spend too much time to figure out if it's a correct fix, but with the following change

```
diff --git a/gradle-conjure/build.gradle b/gradle-conjure/build.gradle
index b091199..4f48f77 100644
--- a/gradle-conjure/build.gradle
+++ b/gradle-conjure/build.gradle
@@ -33,6 +33,8 @@ dependencies {
     annotationProcessor 'org.immutables:value'
     compileOnly 'org.immutables:value::annotations'
 
+    testImplementation "org.junit.jupiter:junit-jupiter:5.7.0"
+    testRuntimeOnly "org.junit.vintage:junit-vintage-engine:5.7.0"
     testImplementation gradleTestKit()
     testImplementation 'com.netflix.nebula:nebula-test'
     testImplementation 'com.squareup.okhttp3:mockwebserver'
@@ -57,3 +59,12 @@ tasks.withType(JavaCompile).configureEach {
         check('PublicConstructorForAbstractClass', CheckSeverity.OFF)
     }
 }
+
+test {
+    description = "Runs JUnit 4 tests."
+    useJUnit()
+    scanForTestClasses = false
+    include(["**/*Tests.class", "**/*Test.class"])
+    exclude(["**/testng/**/*.*", "**/jupiter/**/*.*"])
+    reports.junitXml.destination = file("$buildDir/test-results")
+}
```

there are 2 unrelated tests failing:
ConjurePluginTest. works with checkUnusedDependencies
ConjurePublishTypeScriptTest. publishes generated code

ylee088: ^ Resolved in PR #885

atrocities (approved):

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointVersionBound.java
```diff
@@ -59,11 +62,7 @@ public void setMinVersion(String minVersion) {
     }
 
     public String getMaxVersion() {
```
carterkozak: I think we either want to annotate this `@Nullable`, or return an `Optional<String>` and in the method body use `return Optional.ofNullable(maxVersion);` avoiding the (java) serialization issues around non-serializable `Optional` objects.

Slight preference to return an optional because it's more obvious to API consumers. It will technically be an API break, however beginning to return nulls from a previously non-null method is also a semantic break.

carterkozak (approved):

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/EndpointVersionBound.java
ylee088: Just for future reference, with the return type changed to Optional<String>, I added @JsonInclude(Include.NON_ABSENT) and set .registerModule(new Jdk8Module()) in ObjectMapper. Without these changes the serialization ends up with "max-version":{"empty":false,"present":true}.

PR #898, opened 2021-11-23
Restore optimization

rzpt:
## Before this PR
https://github.com/palantir/gradle-conjure/pull/793 reverted the optimization.

## After this PR
==COMMIT_MSG==
Restore perf optimization to reuse the same classloader for conjure generators when using gradle 7.4.2
==COMMIT_MSG==

## Possible downsides?
None.


rzpt:
@CRogers would you be able to review this?

changelog/@unreleased/pr-898.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: fix
+fix:
+  description: Restore perf optimization to reuse the same classloader for conjure
+    generators
```
CRogers: I lack quite a lot of context here - what makes us think this is safe to re-enable, given the we reverted the previous change and [the underlying gradle bug hasn't been fixed](https://github.com/gradle/gradle/issues/17434)?

Having a look at `ConjureRunnerResource`, it looks like we're using trying to use a `BuildService` to load the classpath as a singleton then interact it with tasks, but are falling foul of thread safety issues? Can we see we can use the [Worker API](https://docs.gradle.org/7.4/userguide/worker_api.html) instead? I think it's designed for pretty much exactly this use case. It [has different isolation modes](https://docs.gradle.org/7.4/userguide/worker_api.html#changing_the_isolation_mode) (`processIsolation`, `classLoaderIsolation` etc) and I'd hope gradle should manage handle the classloader lifetimes etc and is designed to be used concurrently.

carterkozak:
Looks like we can re-enable this behind a check for gradle 7.4.2 which makes service registration thread-safe

rzpt:
Ready for review.

carterkozak (approved):
thanks!

PR #951, opened 2022-02-08
ignore warnings from generated code in eclipse

rzpt:
## Before this PR

Eclipse warns about generated code.

## After this PR

==COMMIT_MSG==
ignore warnings from generated code in eclipse
==COMMIT_MSG==

## Possible downsides?

Strict improvement. The only problem that I know of is gradle-baseline applies the eclipse configuration in afterEvaluate, which I use here, but there is no guarantee that gradle-baseline's code will be applied before this one.

A possible alternate solution is to add `@SuppressWarnings("all")` in the conjure generated code.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -59,6 +59,7 @@ import org.gradle.api.tasks.TaskOutputs;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.language.base.plugins.LifecycleBasePlugin;
 import org.gradle.plugins.ide.eclipse.EclipsePlugin;
+import org.gradle.plugins.ide.eclipse.model.EclipseModel;
 import org.gradle.plugins.ide.idea.IdeaPlugin;
 import org.gradle.plugins.ide.idea.model.IdeaModule;
 
@@ -243,6 +244,8 @@ public final class ConjurePlugin implements Plugin<Project> {
             if (extraConfig != null) {
                 extraConfig.accept(subproj);
             }
+
+            suppressIdeWarnings(subproj);
         });
     }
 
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -600,6 +604,32 @@ static void applyDependencyForIdeTasks(Project project, TaskProvider<?> compileC
...
+                }
+            });
+        });
+    }
+
+    private static void configureEclipseJdt(Properties properties) {
+        properties.put("org.eclipse.jdt.core.compiler.problem.missingOverrideAnnotation", "ignore");
+        properties.put(
+                "org.eclipse.jdt.core.compiler.problem.missingOverrideAnnotationForInterfaceMethodImplementation",
+                "disabled");
```
carterkozak, line 629: Where do we have code with missing override annotations? That sounds like a bug that should be fixed

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -600,6 +604,32 @@ static void applyDependencyForIdeTasks(Project project, TaskProvider<?> compileC
...
+        properties.put("org.eclipse.jdt.core.compiler.problem.unusedPrivateMember", "ignore");
```
carterkozak, line 630: Same here, we can generate `@SuppressWarnings("unused")` inline if we know a default implementation ignores a field.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
rzpt: All methods in a DefaultBuilder seem to be missing it `class DefaultBuilder implements Builder`

carterkozak: Any chance you can file an issue?
https://github.com/palantir/conjure-java/issues

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
rzpt: `_plainSerDe` on Async and Blocking classes.
`UndertowRuntime runtime` on an Endpoint.

PR #959, opened 2022-02-22
use worker api in conjure generation tasks

jochs:
## Before this PR

In process execution of codegen across many files would cause the gradle daemon to run out of metaspace and crash with an OOM.

~Note that in my testing, using a worker with NoIsolation mode did not work, because the parent processes environment was not passed down to the worker. Using process isolation with explicit passing of the env was the only way I could get conjure typescript codegen to work correctly (and it doesn't seem to have incurred that much of a perf hit, which was really my only concern)~

## After this PR

==COMMIT_MSG==
Migrates conjure generation to the gradle worker api to eliminate repeated loading of the same classpath
==COMMIT_MSG==


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java
```diff
@@ -87,27 +92,29 @@ protected File outputDirectoryFor(File _file) {
 
     /** Entry point for the task. */
     public void compileFiles() {
-        getSource().getFiles().forEach(file -> {
-            File thisOutputDirectory = outputDirectoryFor(file);
-
+        Map<String, String> environment = System.getenv();
+        WorkQueue workQueue = getWorkerExecutor().processIsolation(processWorkerSpec -> {
```
CRogers: I need to read up exactly what this does - but if you know already, does this spin up one process per source file or one process for **all** the source files (that is not in the gradle daemon?)

CRogers: https://docs.gradle.org/7.4/userguide/worker_api.html#creating_a_worker_daemon implies to me this will create one worker daemon for all the tasks.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java
```diff
@@ -179,12 +180,13 @@ public void invoke(Project project, String failedTo, List<String> unloggedArgs,
...
         @Override
-        public void invoke(Project project, String failedTo, List<String> unloggedArgs, List<String> loggedArgs) {
-            project.getLogger().info("Running in-process java with args: {}", loggedArgs);
+        public void invoke(
+                ExecOperations _execOperations, String failedTo, List<String> unloggedArgs, List<String> loggedArgs) {
             List<String> combinedArgs = ImmutableList.<String>builder()
                     .addAll(unloggedArgs)
                     .addAll(loggedArgs)
                     .build();
+            log.info("Running in-process java with args: {}", loggedArgs);
```
CRogers, line 160: I'm trying to work out _why_ the metaspace was running out and how we expect running in an external worker daemon to help.

I've read this code and it seems we are reverse engineering the classpath from the conjure generator binary, then running the main method. My guess is that we load the classes multiple times per conjure generator invocation (possible over several gradle runs if the gradle daemon is reused) these classes are not released from memory (?) causing it to run out metaspace.

I suspect that the worker daemons are going to have the same issue here, since we repeatedly classload inside of them and the worker daemons are persisted across gradle invocations, and will also OOM. I'm not sure if this will fail the build or whether gradle will try and run it again, but it feels like we aren't fixing the source of the issue - the memory leak.

I guess this is what #898 is trying to do - reuse the classloader rather than make a new one each time.

I think for both #898 and this issue we might be able to kill two birds with one stone - perhaps we can get gradle to manage the lifecycle of the classpath through the `classloaderIsolation` part of the worker api? Rather than spinning out a process isolated worker daemon that has a bit more metaspace overhead and still does our own classloading?

CRogers: Hmm, I see [we use some bytebuddy to replace the `System.exit` method](https://github.com/palantir/gradle-conjure/blob/ed6aa328cb527399c2e4ade472140dcf2f373c64/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java#L105-L116). @carterkozak do you think it would be possible to still do that with gradle doing the classloading? (and does the above make sense to you?)

CRogers: Actually, I see there's a `toJar` method instead of [loading it directly](https://github.com/palantir/gradle-conjure/blob/ed6aa328cb527399c2e4ade472140dcf2f373c64/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java#L115-L116), so I'm going to guess yes we can still use gradle for this.

carterkozak: That's worth a shot, but I imagine it will be expensive so we'll want to reuse the worker initialization as much as possible.

CRogers: I either hope that gradle will reuse the classpath isolated worker (will need to test) or we can make a root level plugin to hold on to the various worker executor states.

jochs: gradle will reuse the workers as much as possible (or so they say)

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java
jochs: yep that was my understanding!

jochs:
@CRogers I think I need to do some refactoring, but the classpath isolation seems functional! This is really cool tbh

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java
```diff
@@ -111,13 +92,12 @@ static ConjureRunner createNewRunner(File executable) throws IOException {
                                     MemberSubstitution.relaxed()
                                             .method(ElementMatchers.is(System.class.getMethod("exit", int.class)))
                                             .replaceWith(GradleExecStubs.getStubMethod())))
-                    .make(typePool)
-                    .load(classLoader, ClassLoadingStrategy.Default.INJECTION)
+                    .make()
+                    .load(Thread.currentThread().getContextClassLoader(), ClassLoadingStrategy.Default.INJECTION)
```
CRogers: Is this loading it into the current classloader? Won't that mean if this method is called multiple times, we'll load the same patched class multiple times? If so, does this work with main method in the gradle configuration classloader and the generator code in the worker classloader.

We also have a `classLoader` parameter on this method, whose argument is only ever `Thread.currentThread().getContextClassLoader()` - should we use that instead?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java
```diff
@@ -37,70 +34,54 @@
...
+            }
+        }
+        return new ExternalProcessConjureRunner(executable);
     }
 
-    interface ConjureRunner extends Closeable {
-
-        void invoke(Project project, String failedTo, List<String> unloggedArgs, List<String> loggedArgs);
+    static WorkQueue getIsolatedConjureWorkQueue(WorkerExecutor workerExecutor, File executable) {
+        return workerExecutor.classLoaderIsolation(classLoaderWorkerSpec -> conjureExecutableClasspath(executable)
```
CRogers, line 67: I think we need to either write a test or at least manually check that this reuses classloaders for the same conjure executable.

CRogers: Hmm, I suspect they may not be:

https://github.com/gradle/gradle/blob/56445ef7f3b2117631f4afbbd9845b73ec723f82/subprojects/workers/src/main/java/org/gradle/workers/internal/IsolatedClassloaderWorker.java#L52

jochs: Moved back to the daemon because it looks like that will allow you to re-use the underlying classloader: https://github.com/gradle/gradle/blob/ab01e7fc05e4d14186b0a706e0737fe0b20f458f/subprojects/workers/src/main/java/org/gradle/workers/internal/WorkerDaemonServer.java#L98

jochs: That didn't actually work -- although gradle purports to reuse the underlying classloader, I think there's a "bug" where it discards the whole daemon object and recreates it.

Leaving this in the classpath isolation mode - it will still have to construct the classloader a bunch of times but at least the classpaths will be GC'd correctly

tpetracca:
Discussed offline with Burro folks; this won't actually fix their underlying issue.

robert3005:
but it would speed up compilation? Why would it not, can you provide details?

CRogers:
@robert3005 The immediate burro issue was due to gradle ooms due to the recent immutables bug; this does not fix that. The PR as is does not speed up anything as the classloaders are not reused - we'd need to spend time to achieve that when we have other more important immediate tasks on our plate. For most non-monorepo projects reusing classloaders will not speed up compilation significantly; we may come back to this when we have more time to help out for the monorepo use case.

robert3005:
thanks for explanation @CRogers

PR #1014, opened 2022-04-12
Add service and endpoint names and deprecated flag to EndpointVersionBound.

ylee088:
Extra information for more info in detailed reports.


carterkozak (changes requested):
I don't think we should include redundant information in this API. At the limit, we'd end up with the entire IR embedded here. Can we join the results with the current conjure IR to derive this information?

ylee088:
Understood, will look into alternative solutions in conjure-backcompat. Though I feel that adding at least the endpoint name in the EndpointVersionBound object would still be helpful, as we have had instances where http path + method from different IRs clashed.

carterkozak:
> we have had instances where http path + method from different IRs clashed

The service-name and endpoint-name can safely change in every release without causing an API break, the unique identifier must be the http method + normalized http path (in which path variables are replaced with a consistent sentinel value e.g. `/foo/{variable}` becomes `/foo/{}` such that renaming `variable` doesn't break the comparison).

PR #1049, opened 2022-05-17
The conjure IR compiler supports standard customization

carterkozak:
Custom arguments are supported using the same mechanism as
generators, but using the `parser` key:
```groovy
conjure {
  parser {
    verbose = true
  }
}
```

==COMMIT_MSG==
The conjure IR compiler supports standard customization using the `parser` key
==COMMIT_MSG==

## Possible downsides?
`parser` isn't syntax we use elsewhere, however `compiler` and "IR" are not obvious to users, so I think parser is reasonable.


gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/GeneratorOptions.java
```diff
@@ -87,4 +87,28 @@ public GeneratorOptions addFlag(String flag) {
...
+
+        GeneratorOptions that = (GeneratorOptions) other;
+
+        return storage.equals(that.storage);
+    }
+
+    @Override
+    public int hashCode() {
+        return storage.hashCode();
+    }
```
carterkozak, line 108: This class was used as a task input, I don't know how gradle calculates that sort of thing, but I suspect the presence of a reasonable equals/hashCode doesn't hurt. I couldn't find docs on how task inputs are compared, I'd really appreciate if someone could point me in the right direction!

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileIrTask.java
```diff
@@ -88,6 +88,10 @@ public final File getOutputFile() {
     @Optional
     public abstract MapProperty<String, Serializable> getConjureExtensions();
 
+    @Input
+    @Optional
+    public abstract MapProperty<String, Object> getOptions();
```
carterkozak, line 93: In practice I suspect the places using `GeneratorOptions` as an input would benefit from `MapProperty<String, Object>`, but I didn't update existing consumers in this PR.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -104,14 +104,13 @@ public final class ConjurePlugin implements Plugin<Project> {
...
+        ConjureBasePlugin conjureBasePlugin = project.getPlugins().apply(ConjureBasePlugin.class);
 
-        TaskProvider<?> compileIrTask = project.getTasks().named(ConjureBasePlugin.COMPILE_IR_TASK);
+        TaskProvider<?> compileIrTask = conjureBasePlugin.compileIrProvider();
         TaskProvider<GenerateConjureServiceDependenciesTask> serviceDependencyTask = project.getTasks()
                 .named(ConjureBasePlugin.SERVICE_DEPENDENCIES_TASK, GenerateConjureServiceDependenciesTask.class);
 
-        ConjureExtension conjureExtension =
-                project.getExtensions().create(ConjureExtension.EXTENSION_NAME, ConjureExtension.class);
+        ConjureExtension conjureExtension = conjureBasePlugin.conjureExtension();
```
carterkozak, line 113: I moved the `ConjureExtension` registration to `ConjureBasePlugin`, and updated the interacting pieces to pass references from the plugin registration rather than digging through get/find-by-X. It seemed cleaner, but I wouldn't be surprised if it's an unspeakable sin against gradle.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/TestVersions.java
```diff
@@ -19,7 +19,7 @@
 public final class TestVersions {
     private TestVersions() {}
 
-    public static final String CONJURE = "4.11.2";
+    public static final String CONJURE = "4.30.0";
```
carterkozak, line 22: Updated so that the `--verbose` flag is actually used rather than listed as an unmatched input -- in practice it's not necessary for the test to pass, and we're only validating plumbing. I'll revert this out if it causes anything to fail.

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/GeneratorOptions.java
fawind: All inputs need to be serializable. Gradle then serializes inputs to a binary format and compares those binary formats to see if the input has changed (think it compares checksums of the serialized representation but not super sure about the details here).

So hashcode shouldn't be necessary but yeah, doesn't hurt to have one!

carterkozak: Thanks!

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
```diff
@@ -45,6 +46,11 @@ public final void python(@DelegatesTo(GeneratorOptions.class) Closure<GeneratorO
         closure.call();
     }
 
+    public final void parser(@DelegatesTo(GeneratorOptions.class) Closure<GeneratorOptions> closure) {
```
CRogers: You should use `Closure` arguments as raw types in extensions for the gradle integration in intellij to work properly (without making a big warning box). Also, for `Closure<V>`, `V` is the return type so it shouldn't really be the delegate type.

CRogers: Probably a nice thing to make an errorprone rule of actually

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
fawind: Hmm, looks good to me but actually not sure why this isn't applied more often. I guess it requires a tighter coupling between plugins but that seems fine in this case. Maybe @CRogers has a stronger opinion here :)

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
carterkozak: Huh, interesting! I copied from the other methods on this object, I'll update all of 'em.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
```diff
@@ -98,6 +116,9 @@ private static TaskProvider<CompileIrTask> createIrTasks(
             compileIr.getExecutableDir().set(extractCompilerTask.flatMap(ExtractExecutableTask::getOutputDirectory));
             compileIr.getOutputIrFile().set(irDir.map(dir -> dir.file(project.getName() + ".conjure.json")));
             compileIr.getProductDependencies().set(project.provider(pdepsExtension::getProductDependencies));
+            compileIr
+                    .getOptions()
+                    .set(project.provider(() -> conjureExtension.getParser().getProperties()));
```
fawind, line 121: Afaik, general best practice is to make all fields of an extension (i.e. `conjureExtension`) a property and that way you get the lazy wiring out of the box. But this works as well and don't think we need to refactor the extension now.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
CRogers: I have done both. I don't know how I feel about one versus the other. The just access the plugin method feels nicer from a general-java-coding standpoint, everything in Gradle land does the latter though.

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
CRogers: I only just discovered this recently too!

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
carterkozak: That makes sense, though making the change in this PR would result in either a large refactor, or inconsistency within the ConjureExtension -- what do you think?

fawind: Yeah, I would just leave it like this! Just wanted to flag but don't think its a big deal!

CRogers: It definitely is nicer to just hook up the properties together directly, especially as Gradle can then work out the task dependencies automatically if there are files involved in outputs and inputs of different tasks. The one thing I will say is the extra providers makes the stacktrace more understandable when you have some horrible not-enough-laziness bug like I recently encountered with gradle-jdks, but I don't think that's a good enough reason to not just link them together (and you can always add these back when debugging).

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
```diff
@@ -27,26 +27,31 @@ public class ConjureExtension {
 
     private final GeneratorOptions typescriptOptions = new GeneratorOptions();
     private final GeneratorOptions javaOptions = new GeneratorOptions();
+    private final GeneratorOptions parserOptions = new GeneratorOptions();
     private final GeneratorOptions pythonOptions = new GeneratorOptions();
     private final Map<String, GeneratorOptions> genericOptions = new HashMap<>();
 
-    public final void typescript(@DelegatesTo(GeneratorOptions.class) Closure<GeneratorOptions> closure) {
+    public final <T> void typescript(@DelegatesTo(GeneratorOptions.class) Closure<T> closure) {
```
CRogers: Hmm, I wonder if parameterising the value like this works in the gradle integration or it really does have to be a raw type like I've been doing.

CRogers:
:+1:

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureBasePlugin.java
carterkozak: I've filed #1050 to track

gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
carterkozak: Sorry, I misinterpreted  your comment, my fault -- fixed

PR #1066, opened 2022-06-11
gradle task configuration avoidance

rzpt:
## Before this PR
We are unnecessarily creating and configuring tasks during gradle configuration. See https://docs.gradle.org/current/userguide/task_configuration_avoidance.html for more details.

## After this PR
==COMMIT_MSG==
Avoid creating and configuring gradle tasks
==COMMIT_MSG==

## Possible downsides?
None.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -310,7 +310,7 @@ public final class ConjurePlugin implements Plugin<Project> {
                 return;
             }
 
-            proj.getTasks().withType(checkUnusedDependenciesTask, task -> {
+            proj.getTasks().withType(checkUnusedDependenciesTask).configureEach(task -> {
                 try {
                     Method ignoreMethod = task.getClass().getMethod("ignore", String.class, String.class);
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -591,9 +591,10 @@ static void applyDependencyForIdeTasks(Project project, TaskProvider<?> compileC
                     module.getGeneratedSourceDirs(), project.file(JAVA_GENERATED_SOURCE_DIRNAME)));
         });
         project.getPlugins().withType(EclipsePlugin.class, _plugin -> {
-            Task task = project.getTasks().findByName("eclipseClasspath");
-            if (task != null) {
-                task.dependsOn(compileConjure);
+            try {
+                project.getTasks().named("eclipseClasspath").configure(t -> t.dependsOn(compileConjure));
+            } catch (UnknownDomainObjectException e) {
```
CRogers: Kinda frustrating we have to do this, but I can't see another way of doing it with lazy api.

Might be a bit clearer to use [`UnknownTaskException`](https://docs.gradle.org/current/javadoc/org/gradle/api/UnknownTaskException.html) as that's what's explicitly called out in the javadoc.

CRogers:
:+1: given comment

PR #1053, opened 2022-05-24
Improvement: Bring back generateNpmrc with full configurability

robert3005:
## After this PR

==COMMIT_MSG==
Provide unopinionated task for generating npmrc that lets users publish conjure clients to their npm repos
==COMMIT_MSG==

PR #1077, opened 2022-07-20
Don't log npm token response on serde error

fawind:
## Before this PR
FLUP on https://github.com/palantir/gradle-conjure/pull/1053.

## After this PR

The jackson exception message might contain the raw response which can possibly contain sensitive data. E.g. if the registry decides to change their response wire-format, we would log the response including the received publishing token.

==COMMIT_MSG==
Don't log npm token response on serde error
==COMMIT_MSG==


carterkozak (approved):
Thanks!

PR #1089, opened 2022-08-29
Prefer Javax versions instead of the Jakarta Versions

mglazer:
## After this PR

==COMMIT_MSG==
Prefer Javax versions instead of the Jakarta Versions

The jakarta versions were fine, until the Jakarta EE9 releases, where
for example:

javax.ws.rs:javax.ws.rs-api 2.0.0 -> jakarta.ws.rs:jakarta.ws.rs-api 2.1.6

was totally fine. However, once going to:

jakarta.ws.rs:jakarta.ws.rs-api 2.1.6 -> jakarta.ws.rs:jakarta.ws.rs-api 3.0.0

we are no longer able to trivially make that version change.

In order to better handle other plugins like Gradle Consistent Versions,
which lock all modules with the same identifier into the same version,
we have to make this plugin depend on the legacy javax versions of some
APIs, such that can can have both 2.0.0 and 3.0.0 on the classpath at
the same time.
==COMMIT_MSG==

## Possible downsides?

This change effectively drops full support for JDK8 and earlier.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -110,7 +110,6 @@ public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
         ConjurePlugin.addGeneratedToMainSourceSet(project);
 
         project.getDependencies().add("api", "com.palantir.conjure.java:conjure-lib");
-        project.getDependencies().add("compileOnly", ConjurePlugin.ANNOTATION_API);
 
         TaskProvider<WriteGitignoreTask> generateGitIgnore = ConjurePlugin.createWriteGitignoreTask(
                 project, "gitignoreConjure", project.getProjectDir(), ConjurePlugin.JAVA_GITIGNORE_CONTENTS);
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -289,7 +289,7 @@ public final class ConjurePlugin implements Plugin<Project> {
     }
 
     private static void setupJerseyProject(Project project) {
-        project.getDependencies().add("api", "javax.ws.rs:javax.ws.rs-api");
+        project.getDependencies().add("api", "javax.ws.rs:javax.ws.rs-api:2.1.1");
         project.getDependencies().add("compileOnly", ANNOTATION_API);
     }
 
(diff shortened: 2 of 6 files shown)


mglazer:
I recognize this _might_ be a breaking change for this plugin, if someone doesn't have javax.ws.rs:javax.ws.rs-api pinned....maybe I just add the version in the dependency?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -110,7 +110,6 @@ private static void createGenerateTask(
         ConjurePlugin.addGeneratedToMainSourceSet(project);
 
         project.getDependencies().add("api", "com.palantir.conjure.java:conjure-lib");
-        project.getDependencies().add("compileOnly", ConjurePlugin.ANNOTATION_API);
```
mglazer, line 113: this will break Java 9 and earlier...I believe that's fine at this point?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -285,12 +286,21 @@ private static void setupDialogueProject(Project project) {
     private static void setupRetrofitProject(Project project) {
         project.getDependencies().add("api", "com.google.guava:guava");
         project.getDependencies().add("api", "com.squareup.retrofit2:retrofit");
-        project.getDependencies().add("compileOnly", ANNOTATION_API);
+        project.getDependencies().add("compileOnly", "javax.annotation:javax.annotation-api:1.3.2");
```
mglazer, line 289: Do we even support retrofit anymore?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -285,12 +286,21 @@ private static void setupDialogueProject(Project project) {
...
+        if (opts.has(JAKARTA_PACKAGES)) {
+            // version is not set to let the consumer supply it
+            project.getDependencies().add("api", "jakarta.ws.rs:jakarta.ws.rs-api");
+        } else {
+            // version is included explicitly to ensure it doesn't have to be pinned
+            // javax version is chosen to avoid conflicts with jakarta EE9 libraries
+            project.getDependencies().add("api", "javax.ws.rs:javax.ws.rs-api:2.1.1");
+            // make Generated available for Java 8 and earlier, don't do anything with Jakarta (they'll
+            // have a modern JDK if they're migrating to Jakarta)
+            project.getDependencies().add("compileOnly", "javax.annotation:javax.annotation-api:1.3.2");
```
mglazer, line 302: I added this just because it doesn't hurt.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/TestVersions.java
```diff
@@ -20,8 +20,8 @@ public final class TestVersions {
     private TestVersions() {}
 
     public static final String CONJURE = "4.30.0";
-    public static final String CONJURE_JAVA = "5.17.0";
-    public static final String CONJURE_JAVA_DIALOG = "1.50.0";
+    public static final String CONJURE_JAVA = "6.61.0";
```
mglazer, line 23: Bumped so I can actually test the thing I want to test

mglazer:
This was not the right approach.

PR #1123, opened 2022-11-28
conjure-java-local applies pdeps using standard mechanisms

carterkozak:
Note that this won't support per-endpoint dependencies, but reuses shared logic with the standard conjure-java plugin describing product deps.

==COMMIT_MSG==
conjure-java-local applies pdeps using standard mechanisms, adding dependency information correctly to the jar resource in addition to the jar manifest
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -21,8 +21,8 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import com.palantir.conjure.java.serialization.ObjectMappers;
 import com.palantir.gradle.conjure.api.ConjureExtension;
-import com.palantir.gradle.dist.ConfigureProductDependenciesTask;
 import com.palantir.gradle.dist.ProductDependency;
+import com.palantir.gradle.dist.RecommendedProductDependenciesExtension;
 import com.palantir.gradle.dist.RecommendedProductDependenciesPlugin;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPluginIntegrationSpec.groovy
```diff
@@ -180,15 +184,17 @@ class ConjureJavaLocalCodegenPluginIntegrationSpec extends IntegrationSpec {
...
                 '"product-group":"com.palantir.conjure",' +
                 '"product-name":"conjure",' +
                 '"minimum-version":"1.2.0",' +
                 '"recommended-version":"1.2.0",' +
                 '"maximum-version":"2.x.x",' +
                 '"optional":false' +
                 '}]}'
+        def jarFile = file('conjure-api/build/libs/conjure-api-1.0.0.jar')
+        readManifestRecommendedProductDeps(jarFile) == expected
+        readResourceRecommendedProductDeps(jarFile) == expected
```
carterkozak, line 197: Bug is caught here! Previously, this successfully read a file with an empty array of dependencies....

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -118,13 +118,11 @@ private static void createGenerateTask(
...
-                .named("configureProductDependencies", ConfigureProductDependenciesTask.class)
-                .configure(task -> {
-                    task.setProductDependencies(
-                            conjureIrFile.map(ConjureJavaLocalCodegenPlugin::extractProductDependencies));
-                    task.dependsOn(extractConjureIr);
-                });
+
+        project.getExtensions()
+                .getByType(RecommendedProductDependenciesExtension.class)
+                .getRecommendedProductDependenciesProvider()
```
carterkozak, line 124: This provider is used as the source of truth in a couple places, previously we only updated one of its outputs

tpetracca (approved):

PR #1128, opened 2022-12-06
extractIr task dependencies are configured correctly

carterkozak:
==COMMIT_MSG==
extractIr task dependencies are configured correctly
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -31,22 +31,19 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
-import java.util.regex.Pattern;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.file.RegularFile;
 import org.gradle.api.plugins.JavaBasePlugin;
 import org.gradle.api.plugins.JavaLibraryPlugin;
 import org.gradle.api.provider.Provider;
-import org.gradle.api.tasks.Copy;
 import org.gradle.api.tasks.TaskProvider;
 
 public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
     private static final ObjectMapper OBJECT_MAPPER = ObjectMappers.newClientObjectMapper();
     private static final String CONJURE_CONFIGURATION = "conjure";
-    private static final Pattern DEFINITION_NAME =
-            Pattern.compile("(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?)(\\.conjure)?\\.json");
 
     @Override
     public void apply(Project project) {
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractConjureIrTask.java
```diff
@@ -0,0 +1,49 @@
...
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.file.Directory;
+import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.OutputDirectory;
+import org.gradle.api.tasks.Sync;
+
+public abstract class ExtractConjureIrTask extends Sync {
```
CRogers, line 27: This is definitely my preferred way of writing all Gradle tasks in plugins now, making an implementation of the standard task type with real input/output properties.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractConjureIrTask.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final Pattern DEFINITION_NAME =
+            Pattern.compile("(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?)(\\.conjure)?\\.json");
+
+    public ExtractConjureIrTask() {
+        rename(DEFINITION_NAME, "$1.conjure.json");
+        from(getIrConfiguration());
+        Provider<Directory> outputDirectory =
+                getProject().getLayout().getBuildDirectory().dir("conjure-ir");
+        into(outputDirectory);
+        getConjureIr().set(outputDirectory);
```
CRogers: This is a bit funky to me, if someone changed the `getConjureIr()` value then the `into` value would not get updated. Something like

```java
getConjureIr().set(getProject().getLayout().getBuildDirectory().dir("conjure-ir"));
into(getConjureIr())
```

seems a bit safer to me.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -117,8 +113,9 @@ private static void createGenerateTask(
         TaskProvider<WriteGitignoreTask> generateGitIgnore = ConjurePlugin.createWriteGitignoreTask(
                 project, "gitignoreConjure", project.getProjectDir(), ConjurePlugin.JAVA_GITIGNORE_CONTENTS);
 
-        Provider<File> conjureIrFile = extractConjureIr.map(
-                irTask -> new File(irTask.getDestinationDir(), project.getName() + ".conjure.json"));
+        Provider<File> conjureIrFile = extractConjureIr
+                .flatMap(task -> task.getConjureIr().file(project.getName() + ".conjure.json"))
```
CRogers: Have you tested this sets up the task dependency correctly (you can run with `--dry-run`)? This seems functionally the same to me as the previous code, except with some added custom types.

carterkozak: womp:
```
Querying the mapped value of map(flatmap(provider(task 'extractConjureIr', class com.palantir.gradle.conjure.ExtractConjureIrTask))) before task ':remote-api:extractConjureIr' has completed is not supported
```

carterkozak:
Whew, so I've updated this PR based on feedback, and I believe it _does_ make the code easier to follow, and probably gives gradle more information to help us. However, I don't believe any change is necessary in this project as the root cause appears to be resolved by: https://github.com/palantir/sls-packaging/pull/1417

PR #1155, opened 2023-02-08
consider 'useJakartaNamespaces' option when setting up dependencies

bjlaub:
## Before this PR
Previously, ConjurePlugin would apply only "jakarta" dependencies to java subprojects, but this allows projects that are using Jakarta dependencies for which resolution has selected a higher version that has the newer "jakarta" namespace to possibly be inconsistent with the "useJakartaNamespaces" flag (for instance, if the flag were set to false, it would generate code that would not work, and possibly vice-versa).

## After this PR

==COMMIT_MSG==
by default gradle-conjure will generate a project with the older "javax" dependencies unless "useJakartaNamespaces" is set to true, and then it will select the newer "jakarta" dependencies.
==COMMIT_MSG==

## Possible downsides?
It's possible that a project which does not use https://github.com/palantir/jakarta-package-alignment will end up in a state where other projects use jakarta libs, but generated conjure projects use javax libs.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -281,8 +281,7 @@ public final class ConjurePlugin implements Plugin<Project> {
     }
 
     private static void setupRetrofitProject(Project project, Supplier<GeneratorOptions> optionsSupplier) {
-        boolean useJakarta = optionsSupplier.get().has("useJakartaNamespaces")
-                && ((Boolean) optionsSupplier.get().get("useJakartaNamespaces"));
+        boolean useJakarta = Dependencies.isUseJakartaNamespaces(optionsSupplier.get());
         project.getDependencies().add("api", "com.google.guava:guava");
         project.getDependencies().add("api", "com.squareup.retrofit2:retrofit");
         project.getDependencies()
@@ -292,8 +291,7 @@ public final class ConjurePlugin implements Plugin<Project> {
     }
 
     private static void setupJerseyProject(Project project, Supplier<GeneratorOptions> optionsSupplier) {
-        boolean useJakarta = optionsSupplier.get().has("useJakartaNamespaces")
-                && ((Boolean) optionsSupplier.get().get("useJakartaNamespaces"));
+        boolean useJakarta = Dependencies.isUseJakartaNamespaces(optionsSupplier.get());
         project.getDependencies()
                 .add("api", useJakarta ? Dependencies.JAXRS_API_JAKARTA : Dependencies.JAXRS_API_JAVAX);
         project.getDependencies()
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -35,7 +35,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.Consumer;
+import java.util.function.BiConsumer;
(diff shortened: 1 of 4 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -270,27 +271,38 @@ private static boolean shouldConfigureJavaServices(Project project) {
...
+                        "compileOnly",
+                        useJakarta ? Dependencies.ANNOTATION_API_JAKARTA : Dependencies.ANNOTATION_API_JAVAX);
     }
 
-    private static void setupJerseyProject(Project project) {
-        project.getDependencies().add("api", Dependencies.JAXRS_API);
-        project.getDependencies().add("compileOnly", Dependencies.ANNOTATION_API);
+    private static void setupJerseyProject(Project project, Supplier<GeneratorOptions> optionsSupplier) {
+        boolean useJakarta = optionsSupplier.get().has("useJakartaNamespaces")
+                && ((Boolean) optionsSupplier.get().get("useJakartaNamespaces"));
```
carterkozak: let's either extract a constant, or assuming this returns null when a value  doesn't exist, something like `Boolean.TRUE.equals(optionsSupplier.get().get("useJakartaNamespaces"))`

gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
```diff
@@ -19,12 +19,15 @@
...
+    static final String ANNOTATION_API_JAKARTA = "jakarta.annotation:jakarta.annotation-api:1.3.5";
+
+    static final String ANNOTATION_API_JAVAX = "javax.annotation:javax.annotation-api:1.3.2";
 
     static final String CONJURE_JAVA_LIB = "com.palantir.conjure.java:conjure-lib";
     static final String CONJURE_UNDERTOW_LIB = "com.palantir.conjure.java:conjure-undertow-lib";
     static final String DIALOGUE_TARGET = "com.palantir.dialogue:dialogue-target";
-    static final String JAXRS_API = "jakarta.ws.rs:jakarta.ws.rs-api";
+    static final String JAXRS_API_JAKARTA = "jakarta.ws.rs:jakarta.ws.rs-api";
+    static final String JAXRS_API_JAVAX = "javax.ws.rs:javax.ws.rs-api";
```
carterkozak: Let's specify versions for these, otherwise builds may fail if no constraints exist for javax.ws.rs...etc. Otherwise the jakarta segmentation plugin must be applied first.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
bjlaub: `.get(...)` throws when a value doesn't exist :(

gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
bjlaub: fixed

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
bjlaub: how do you feel about https://github.com/palantir/gradle-conjure/pull/1155/commits/427fd95357866750caea9ad0b916075d09bf0a58#diff-57a6e01c74489ed759713fa386614f0d7bd705d8cba6c7dfaa7f27f8baa9a4e9R38-R43?

carterkozak: lgtm

carterkozak (approved):

PR #1189, opened 2023-03-17
wrap setupDerivedProjects in afterEvaluate block

bjlaub:
without this, in some cases it seems like sometimes the ConjureExtension object is not actually populated with values parsed from the build script at the time that we need them to add dependencies on subprojects. This is problematic when e.g. the "jakartaPackages" flag is set, because it changes the set of dependencies we need to add.


carterkozak:
Merging to unblock this -- we can revisit with a better solution later

carterkozak (approved):

PR #1250, opened 2023-07-11
set the jakartaPackages flag via a gradle property

bjlaub:
## Before this PR
This flag was previously set thusly in a build script:
```
conjure {
    java {
        jakartaPackages = true
    }
}
```

While that's still possible, it potentially causes problems when configuring subprojects due to evaluation order; the flag's value is not populated within the extension object until some time _after_ the plugin application finishes, which results in incorrect dependencies getting added to some subprojects when they are set up in `ConjurePlugin` and similar plugin applications.

## After this PR

==COMMIT_MSG==
Adding a gradle property `conjure.java.jakartaPackages=true` will cause the value to be set eagerly on the extension object such that it will always exist at the time the `ConjurePlugin`'s application logic runs, and so should have the value set correctly at the time subproject dependencies are added.
==COMMIT_MSG==

## Possible downsides?
It's an extra level of annoyance to have to set this e.g. via a `gradle.properties` file.

### DIFF
--- a/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
+++ b/gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
@@ -20,6 +20,7 @@ import groovy.lang.Closure;
 import groovy.lang.DelegatesTo;
 import java.util.HashMap;
 import java.util.Map;
+import org.gradle.api.Project;
 
 public class ConjureExtension {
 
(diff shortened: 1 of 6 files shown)


gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureExtension.java
```diff
@@ -31,12 +32,16 @@ public class ConjureExtension {
     private final GeneratorOptions pythonOptions = new GeneratorOptions();
     private final Map<String, GeneratorOptions> genericOptions = new HashMap<>();
 
-    public ConjureExtension() {
+    public ConjureExtension(Project project) {
```
bjlaub, line 35: we could potentially change this to be `String jakartaPackagesProperty` and just make the caller use `project.findProperty` - i'm not really sure which is preferred.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -523,13 +523,9 @@ class ConjurePluginTest extends IntegrationSpec {
...
-        file('api/build.gradle') << '''
-        conjure {
-            java {
-                jakartaPackages = true
-            }
-        }
-        '''.stripIndent()
+        createFile('gradle.properties') << """
+        conjure.java.jakartaPackages=true
+        """.stripIndent()
```
carterkozak, line 528: might as well test both

PR #1349, opened 2024-02-08
Run `compileTypescript` and `compileConjure` as a dependency of `build`

CRogers:
## Before this PR
We had an issue internally where some repos started failing at publish time because we hit the Gradle 8 task dependency check. The problem was, `compileTypescript` wasn't actually run pre-merge on the PR build, so the excavator merged even though this task would fail. This means the task dependency error did not appear until publish time, which is terrible.

## After this PR
==COMMIT_MSG==
Run `compileTypescript` and `compileConjure` as a dependency of `build` to catch more failure at PR time rather than publish time.
==COMMIT_MSG==

## Possible downsides?
There could be projects that have typescript in them but do not publish it or actually build it anywhere currently. They will be blocked from taking gradle-conjure upgrades as their builds will fail. Ideally they should actually remove typescript or fix it and publish it. We should be able to find whatever error messages are caused by this and track them down.


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -694,4 +701,22 @@ private static Project findDerivedProject(Project project, String projectName) {
             return project.getRootProject().findProject(projectName);
         }
     }
+
+    private static void buildDependsOn(Project project, TaskProvider<?> task) {
```
felixdesouza, line 704: I feel like I've this around a bit, should we move this to `gradle-utils`?

CRogers: Yeah, I had that thought - it could even live on EnvironmentVariables (the circle node == 0 or local bit)

CRogers: https://github.com/palantir/gradle-utils/pull/117

CRogers: Updated to use that version

felixdesouza:
👍

PR #1354, opened 2024-02-14
fix the pattern used to parse local conjure def names

Blizzara:
to allow non-tagged/non-rc'd builds from commits

## Before this PR

Would fail to parse locally published packages with commit-based versions, like `test-api-0.123.0-123-g1a2c3d.conjure.json`

## After this PR

==COMMIT_MSG==
Support packages with commit-based versions for conjure local generation
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTask.java
@@ -27,7 +27,7 @@ import java.util.regex.Pattern;
 public abstract class ConjureLocalGenerateGenericTask extends ConjureLocalGenerateTask {
 
     private static final Pattern PATTERN =
-            Pattern.compile("^(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-g[a-f0-9]+)?)(?:.conjure)?.json$");
+            Pattern.compile("^(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[1-9][0-9]*-g[a-f0-9]+)?)(?:.conjure)?.json$");
 
     @Override
     protected final Map<String, Supplier<Object>> requiredOptions(File irFile) {
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTask.java
@@ -27,7 +27,7 @@ import java.util.regex.Pattern;
 
 public abstract class ConjureLocalGenerateGenericTask extends ConjureLocalGenerateTask {
 
-    private static final Pattern PATTERN = Pattern.compile("^([^.]+)-(.+?)(?:\\.conjure)?\\.json$");
+    private static final Pattern PATTERN = Pattern.compile("^([^.]+)-(.+?)(\\.conjure)?\\.json$");
 
     @Override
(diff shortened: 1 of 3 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTask.java
```diff
@@ -27,7 +27,7 @@
 public abstract class ConjureLocalGenerateGenericTask extends ConjureLocalGenerateTask {
 
     private static final Pattern PATTERN =
-            Pattern.compile("^(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-g[a-f0-9]+)?)(?:.conjure)?.json$");
+            Pattern.compile("^(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-[1-9][0-9]*-g[a-f0-9]+)?)(?:.conjure)?.json$");
```
pkoenig10: Might be better to simplify this pattern to just extract the version then run it through `OrderableSlsVersion.safeValueOf` from https://github.com/palantir/sls-version-java. That way we don't have to redefine the version regexes here.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTaskTest.groovy
```diff
@@ -31,11 +31,10 @@ class ConjureLocalGenerateGenericTaskTest extends Specification {
         where:
         fileName                           | expectedProductName | expectedProductVersion
         'foo-1.0.0.json'                   | "foo"               | '1.0.0'
-        "foo-baz-1.0.0.json"               | "foo-baz"           | "1.0.0"
+        "foo-baz-2-1.0.0.json"             | "foo-baz-2"         | "1.0.0"
         "foo-1.0.0.conjure.json"           | "foo"               | "1.0.0"
         "foo-1.0.0-rc1.conjure.json"       | "foo"               | "1.0.0-rc1"
-        "foo-1.0.0-gabcd.conjure.json"     | "foo"               | "1.0.0-gabcd"
-        "foo-1.0.0-rc1-gabcd.conjure.json" | "foo"               | "1.0.0-rc1-gabcd"
```
Blizzara, line 38: these two formats were _not_ valid according to SlsVersion

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateGenericTask.java
```diff
@@ -26,8 +27,7 @@
 
 public abstract class ConjureLocalGenerateGenericTask extends ConjureLocalGenerateTask {
 
-    private static final Pattern PATTERN =
-            Pattern.compile("^(.*)-([0-9]+\\.[0-9]+\\.[0-9]+(?:-rc[0-9]+)?(?:-g[a-f0-9]+)?)(?:.conjure)?.json$");
+    private static final Pattern PATTERN = Pattern.compile("^([^.]+)-(.+?)(?:\\.conjure)?\\.json$");
```
pkoenig10:
```suggestion
    private static final Pattern PATTERN = Pattern.compile("^([^.]+)-(.+)(\\.conjure)?\\.json$");
```

Blizzara: The version part needs to be non-greedy or it eats the possibly ".conjure" into itself. 
704bbeb6dfb4e2fc58cf76eac1676ac3229307e5 drops the non-capturing

pkoenig10 (approved):

PR #1401, opened 2024-03-21
Gradle 8 fixes

felixdesouza:
## Before this PR
The way `gradle-conjure` generates files means that it will run into [implicit dependency errors](https://docs.gradle.org/8.5/userguide/validation_problems.html#implicit_dependency). This was happening in [conjure](https://app.circleci.com/pipelines/github/palantir/conjure/2899/workflows/9fc8ef68-982f-4200-ab35-0f48b527887c/jobs/15251/steps).

## After this PR

==COMMIT_MSG==
Set up correct dependencies for Gradle 8.
==COMMIT_MSG==

Key points:
* Instead of using a static directory, use the output of the generator task for the source set
* Set the output to be different for each generator task (java):
  * so instead of `src/generated/java`, we have `build/generated/sources/conjure-<type>/main/java`
  * This mirrors what gradle do with annotation processors: `build/generated/sources/annotationProcessor/java/main`
* Separate out the ide tasks so that running `./gradlew idea` or `./gradlew eclipse` depends on the umbrella task, but each subproject task 

## Possible downsides?
Hard to migrate python/typescript/other sources, as I don't know how they are consumed and that would be a break. In fairness the existing code had set up idea to do the wrong thing to them *anyway* but I imagine people will want to keep stuff in source. This *may* not fix us for the implicit dependency if we're running into the python/ts errors in

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
@@ -85,7 +85,7 @@ public final class ConjurePlugin implements Plugin<Project> {
     static final ImmutableSet<String> JAVA_PROJECT_SUFFIXES = ImmutableSet.of(
             JAVA_DIALOGUE_SUFFIX, JAVA_OBJECTS_SUFFIX, JAVA_JERSEY_SUFFIX, JAVA_RETROFIT_SUFFIX, JAVA_UNDERTOW_SUFFIX);
     static final String JAVA_GENERATED_SOURCE_DIRNAME = "src/generated/java";
-    static final String JAVA_GITIGNORE_CONTENTS = "/src/generated/java/\n";
+    static final String JAVA_GITIGNORE_CONTENTS = "/build/generated/\n";
 
     private static final ImmutableSet<String> FIRST_CLASS_GENERATOR_PROJECT_NAMES = ImmutableSet.<String>builder()
             .addAll(JAVA_PROJECT_SUFFIXES)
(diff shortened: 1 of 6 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -634,14 +629,32 @@ static void applyDependencyForIdeTasks(Project project, TaskProvider<?> compileC
...
             }
+        });
+    }
+
+    static void configureIdeGeneratedSources(Project project, Provider<Directory> generated) {
+        project.getPlugins().withType(IdeaPlugin.class, plugin -> {
+            IdeaModule module = plugin.getModel().getModule();
+
+            // module.getSourceDirs / getGeneratedSourceDirs could be an immutable set, so defensively copy
+            module.setSourceDirs(mutableSetWithExtraEntry(
```
CRogers, line 640: `mutableSetWithExtraEntry` is bringing back memories

felixdesouza: bad memories I hope

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -663,12 +684,10 @@ private static <T> Set<T> mutableSetWithExtraEntry(Set<T> set, T extraItem) {
 
     static TaskProvider<WriteGitignoreTask> createWriteGitignoreTask(
```
CRogers, line 685: These write gitignore tasks are also causing task dep validation errors. Maybe not this one specifically, but there are some more internally that also have problems as they get included into source jars without the dependsOn

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
CRogers: yes, how did you guess??

CRogers (approved):

PR #1403, opened 2024-03-21
Follow conventions for generated directory

pkoenig10:
Small follow up from https://github.com/palantir/gradle-conjure/pull/1401.

## Before this PR

Annotation processors generate code into `build/generated/sources/<name>/java/main` but this plugin generates code into `build/generated/sources/<name>/main/java`.

While this is different than user-authored code, we should strive to follow the directory conventions for generated code.

## After this PR

This plugin generates code into `build/generated/sources/<name>/main/java`.


felixdesouza:
👍  thanks!

PR #1404, opened 2024-03-21
Don't generate gitignore for Java projects

pkoenig10:
These `.gitignore` files are no longer necessary after https://github.com/palantir/gradle-conjure/pull/1401.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -133,9 +133,6 @@ public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
             TaskProvider<Copy> extractConjureIr) {
         ConjurePlugin.ignoreFromCheckUnusedDependencies(project);
 
-        TaskProvider<WriteGitignoreTask> generateGitIgnore = ConjurePlugin.createWriteGitignoreTask(
-                project, "gitignoreConjure", project.getProjectDir(), "/build/generated/\n");
-
         Provider<File> conjureIrFile = extractConjureIr.map(
                 irTask -> new File(irTask.getDestinationDir(), project.getName() + ".conjure.json"));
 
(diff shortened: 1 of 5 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -134,7 +134,7 @@ private static void createGenerateTask(
         ConjurePlugin.ignoreFromCheckUnusedDependencies(project);
 
         TaskProvider<WriteGitignoreTask> generateGitIgnore = ConjurePlugin.createWriteGitignoreTask(
-                project, "gitignoreConjure", project.getProjectDir(), ConjurePlugin.JAVA_GITIGNORE_CONTENTS);
+                project, "gitignoreConjure", project.getProjectDir(), "/build/generated/\n");
```
felixdesouza: I think this is still undesirable since it's generated in the wrong place?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -159,7 +159,8 @@ private static void createGenerateTask(
                                 sanitizePackageName(project.getGroup().toString()));
                         return properties;
                     }));
-                    task.getOutputDirectory().set(project.file(ConjurePlugin.JAVA_GENERATED_SOURCE_DIRNAME));
+                    task.getOutputDirectory()
+                            .set(project.getLayout().getProjectDirectory().dir("src/generated/java"));
```
felixdesouza: I completely forgot this output directory from #1401, I should have moved this over to the new build directory

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
felixdesouza: if this goes back to `src/generated/\n`, I'll fix up the this task so we can actually get rid of it

pkoenig10: Ah thanks, good catch

pkoenig10: I'm happy to fix this, it should just go in `/build/generated` right?

felixdesouza: alright thanks! yeah, should be the same as the other java tasks

felixdesouza:
👍  thank you!

PR #1416, opened 2024-04-10
Correctly parse product name out of filename when producing local conjure

Blizzara:
## Before this PR
At least producing rust code would fail when version is something else than a "1.2.3", ie RCs and commits, producing the generated code into a folder like "product-1.2.3" (for "product-1.2.3-rc1") instead of "product". 

## After this PR

==COMMIT_MSG==
Conjure-local generates code into correct directory also when called on RC- and other more complex versions
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
@@ -17,20 +17,19 @@
 package com.palantir.gradle.conjure;
 
 import com.palantir.sls.versions.OrderableSlsVersion;
-import groovy.lang.Tuple;
-import groovy.lang.Tuple2;
 import java.io.File;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.gradle.api.tasks.CacheableTask;
+import org.immutables.value.Value.Immutable;
 
 @CacheableTask
 public abstract class ConjureLocalGenerateTask extends ConjureGeneratorTask {
 
     protected static final Pattern PATTERN = Pattern.compile("^([^.]+)-(.+?)(\\.conjure)?\\.json$");
 
-    static Tuple2<String, OrderableSlsVersion> parseProductNameAndVersion(String filename) {
+    static ProductNameAndVersion parseProductNameAndVersion(String filename) {
         Matcher matcher = PATTERN.matcher(filename);
         if (!matcher.matches() || matcher.groupCount() < 2) {
             throw new RuntimeException(String.format("Unable to parse conjure dependency name %s", filename));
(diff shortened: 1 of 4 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
```diff
@@ -16,17 +16,41 @@
...
+        }
+        return Tuple.tuple(irName, maybeIrVersion.get());
+    }
+
     @Override
     protected final File outputDirectoryFor(File file) {
         // Strip extension and version
         return getOutputDirectory()
-                .dir(file.getName().substring(0, file.getName().lastIndexOf("-")))
+                .dir(parseProductNameAndVersion(file.getName()).getV1())
```
fawind: This is the actual bug-fix, right?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
```diff
@@ -16,17 +16,41 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.gradle.api.tasks.CacheableTask;
 
 @CacheableTask
 public abstract class ConjureLocalGenerateTask extends ConjureGeneratorTask {
 
+    protected static final Pattern PATTERN = Pattern.compile("^([^.]+)-(.+?)(\\.conjure)?\\.json$");
+
+    static Tuple2<String, OrderableSlsVersion> parseProductNameAndVersion(String filename) {
```
fawind: Nit: Instead of returning a groovy tuple, should we just define a small named wrapper within this class? I.e. a package-private immutables one? That way typing is a bit more explicit.

Blizzara: Yeah, can do - I just avoided that as I dislike the boilerplate it gives 😅  I guess we cannot use java 17 and records here yet?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
Blizzara: yes!

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureLocalGenerateTask.java
fawind: We're very close! But Gradle plugins still need to stay on j11 for a little longer 🤞

Blizzara: ack! switched to immutables in https://github.com/palantir/gradle-conjure/pull/1416/commits/7ea5012e14974d4aeffb2aac1fb3e47b245b51b7

fawind (approved):
Thanks!

PR #1440, opened 2024-05-17
Include versions in all dependencies

pkoenig10:
This ensures that consumer get functioning dependencies when we start relying on new feature of these libraries.

I've just used the latest versions of `conjure-java` and `dialogue`, rather than trying to figure out the minimum required version of these dependencies.

We removed support for Retrofit clients in https://github.com/palantir/conjure-java/pull/2135, so that is no longer tested.

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
@@ -26,9 +26,9 @@ final class Dependencies {
     static final String JAXRS_API_JAKARTA = "jakarta.ws.rs:jakarta.ws.rs-api:3.0.0";
     static final String JAXRS_API_JAVAX = "javax.ws.rs:javax.ws.rs-api:2.1.1";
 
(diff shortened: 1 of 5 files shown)


pkoenig10:
These dependencies are needed by `conjure-java` so we don't actually know the correct version to recommend here.

Closing in favor of https://github.com/palantir/gradle-conjure/pull/1441.

pkoenig10:
Although there is already some coupling here. The `gradle-conjure` plugin is the thing that adds these dependencies. By the same argument, it can't really know what dependencies are needed, but we accept this separation of responsibility as a necessary limitation. It feels a bit weird for something to recommend dependencies but not a version.

pkoenig10:
I've rebased this on top of https://github.com/palantir/gradle-conjure/pull/1441.

pkoenig10:
See https://g.p.b/excavator-checks/excavator-gradle/pull/5064 for more context.

carterkozak (dismissed):

carterkozak (approved):

PR #1461, opened 2024-06-26
generateNpmrc task does not depend on running compileTypeScript first

schlosna:
## Before this PR

Some projects need to use an alternate NPM registry for installing dependencies, however; the `.npmrc` was not generated until after attempting to `installTypeScriptDependencies`

https://github.com/palantir/gradle-conjure/pull/1349/ made this issue fail earlier, blocking `gradle-conjure` upgrades beyond [5.42.0](https://github.com/palantir/gradle-conjure/releases/tag/5.42.0).

```
Process 'command 'npm'' finished with non-zero exit value 1

* Causal chain is:
	org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':foo-api:installTypeScriptDependencies'.
	org.gradle.process.internal.ExecException: Process 'command 'npm'' finished with non-zero exit value 1

* Full exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':foo-api:installTypeScriptDependencies'.
```

As an attempted workaround, if one tried the following:
```
generateNpmrc.registryUri = 'https://artifacts.example.com/npm/'
tasks.installTypeScriptDependencies.dependsOn tasks.generateNpmrc
```

This fails with a circular dependency.
```
FAILURE: Build failed with an exception.

* What went wrong:
Circular dependency between the following tasks:
:foo-api:compileTypeScript
\--- :foo-api:installTypeScriptDependencies
     \--- :foo-api:generateNpmrc
          \--- :foo-api:compileTypeScript (*)

(*) - details omitted (listed previously)
```

## After this PR

==COMMIT_MSG==
generateNpmrc task does not depend on running compileTypeScript first
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1461.v2.yml
+++ b/changelog/@unreleased/pr-1461.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Generate .npmrc before installing TypeScript dependencies
+  links:
+  - https://github.com/palantir/gradle-conjure/pull/1461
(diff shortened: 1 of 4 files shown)


CRogers (approved):
Thanks!

PR #1567, opened 2025-01-09
Remove unnecessary task dependencies

pkoenig10:
Motivated by internal discussions about removing `compileConjureTypeScript` and `installTypeScriptDependencies` from the build task graph when the outputs aren't actually needed.

Admittedly, it is a bit weird that these projects won't be build as part of the lifecycle tasks. But even if we want that, `build` is the wrong task to use - these should be dependencies of `assemble`. The subprojects will already be part of `assemble` as a result of any language-specific plugins applied to those subproject (ie. the Java plugin).

This PR also remove the IDE task dependencies. Opening projects using the IDEA plugin is deprecated and no longer used, so we don't need these tasks. We do however still need to declared the source directories as generated source directories in IntelliJ.

### DIFF
--- a/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
+++ b/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
@@ -656,62 +656,6 @@ class ConjurePluginTest extends IntegrationSpec {
         'peer'     | ''
     }
 
-    def 'sets up idea source sets correctly'() {
-        given:
-        createFile('api/api-jersey/some-extra-source-folder')
-
-        file('build.gradle') << '''
-        subprojects {
-            apply plugin: 'idea'
-
-            idea {
-                module {
-                    sourceDirs += file('some-extra-source-folder')
-                }
-            }
-        }
-        '''.stripIndent()
-
-        when:
-        runTasksSuccessfully('idea')
-
-        then:
-        def slurper = new XmlParser()
-        def module = slurper.parse(file('api/api-jersey/api-jersey.iml'))
-        def sourcesFolderUrls = module.component.content.sourceFolder.@url
-
-        sourcesFolderUrls.size() == 2
-        sourcesFolderUrls.contains('file://$MODULE_DIR$/some-extra-source-folder')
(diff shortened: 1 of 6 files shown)


gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -656,62 +656,6 @@ class ConjurePluginTest extends IntegrationSpec {
...
-    }
-
-    def 'compileTypeScript is run on build for circle node 0'() {
-        when:
-        def stdout = runTasksSuccessfully('build', '--dry-run',
-                '-P__TESTING_CIRCLE_NODE_INDEX=0').standardOutput
-
-        then:
-        stdout.contains ':api:compileTypeScript SKIPPED'
-    }
```
CRogers: I'm pretty sure we want to keep this behaviour. More details at https://github.com/palantir/gradle-conjure/pull/1349, but when rolling out Gradle 8 we kept breaking publish, so we want to make sure that all tasks that get run at publish time happen as a dependency of `build`, to ensure they get tested out pre-merge. Otherwise you get excavators that merge but then break publish.

CRogers:
> But even if we want that, build is the wrong task to use - these should be dependencies of assemble

In reality, it doesn't matter as `build` depends on `assemble` and none of our circle templates call `assemble` directly - everything should call `build` at some point pre-merge.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
pkoenig10: Do we think the `isCircleNode0OrLocal` filtering is really necessary? It's sort of strange for task dependencies to be different on different CI nodes.

I can't imagine the compiling TypeScript is that time consuming. Do we have repos where this requires a meaningful amount of CI resources?

CRogers: Maybe the time isn't too long, but iirc, with gradle-failure-reports, as circle collates data from every job into one list, you can also get the same error multiple times, which isn't very helpful (although unsure if this is still a problem with Circle 4).

CRogers (approved):

PR #1637, opened 2025-06-12
Make `gradle-conjure` configuration cacheable for the `classes` task

FinlayRJW:
## Before this PR
`gradle-conjure` had a range of issues that caused configuration cache to fail for the `classes` task: 
- `ConjureGeneratorTask` and `CompileIrTask` use `getProject()` from within a task action which is not allowed in configuration cache. They do this in two places firstly in `getProject().mkdir(...);` calls and secondly they pass `getProject()` into `GradleExecUtils` which is later passed into and captured by `ConjureRunner`

- `ConjureRunner ` use the `Project` object - the below errors root cause is `ConjureRunner ` even though the don't directly point to it
```
    - task `:log-receiver-api:compileIr` of type `com.palantir.gradle.conjure.CompileIrTask`
        - Exception at `com.palantir.gradle.conjure.CompileIrTask.generate(CompileIrTask.java:112)`
org.gradle.api.InvalidUserCodeException: Invocation of 'Task.project' by task ':log-receiver-api:compileIr' at execution time is unsupported.
	at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
	at com.palantir.gradle.conjure.CompileIrTask.generate(CompileIrTask.java:112)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
```
        - Exception at `com.palantir.gradle.conjure.ConjureGeneratorTask.lambda$compileFiles$0(ConjureGeneratorTask.java:105)`
org.gradle.api.InvalidUserCodeException: Invocation of 'Task.project' by task ':log-receiver-api:compileConjureDialogue' at execution time is unsupported.
	at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
	at com.palantir.gradle.conjure.ConjureGeneratorTask.lambda$compileFiles$0(ConjureGeneratorTask.java:105)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at com.palantir.gradle.conjure.ConjureGeneratorTask.compileFiles(ConjureGeneratorTask.java:90)
	at com.palantir.gradle.conjure.ConjureGeneratorTask$1.execute(ConjureGeneratorTask.java:52)
	at com.palantir.gradle.conjure.ConjureGeneratorTask$1.execute(ConjureGeneratorTask.java:49)
	at java.base/java.lang.Thread.run(Thread.java:1583)
        - Exception at `com.palantir.gradle.conjure.ConjureGeneratorTask.lambda$compileFiles$0(ConjureGeneratorTask.java:99)`
org.gradle.api.InvalidUserCodeException: Invocation of 'Task.project' by task ':log-receiver-api:compileConjureDialogue' at execution time is unsupported.
	at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
	at com.palantir.gradle.conjure.ConjureGeneratorTask.lambda$compileFiles$0(ConjureGeneratorTask.java:99)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at com.palantir.gradle.conjure.ConjureGeneratorTask.compileFiles(ConjureGeneratorTask.java:90)
	at com.palantir.gradle.conjure.ConjureGeneratorTask$1.execute(ConjureGeneratorTask.java:52)
	at com.palantir.gradle.conjure.ConjureGeneratorTask$1.execute(ConjureGeneratorTask.java:49)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
- `ExtractExecutableTask` use `getProject` in `doFirst`
```
task `:extractConjure` of type `com.palantir.gradle.conjure.ExtractExecutableTask`
        - Exception at `com.palantir.gradle.conjure.ExtractExecutableTask$1.execute(ExtractExecutableTask.java:66)`
org.gradle.api.InvalidUserCodeException: Invocation of 'Task.project' by task ':extractConjure' at execution time is unsupported.
	at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
	at com.palantir.gradle.conjure.ExtractExecutableTask$1.execute(ExtractExecutableTask.java:66)
	at com.palantir.gradle.conjure.ExtractExecutableTask$1.execute(ExtractExecutableTask.java:62)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

## After this PR
- in `GradleExecUtils` we now pass in `ExecOperations` and `BuildServiceRegistry` rather than using `Project` - these are injected in `ConjureGeneratorTask` and `ConjureJavaLocalGeneratorTask` which has been converted to an abstract class, this then feeds into `ConjureRunner` and we can remove the use of `Project`
- in `ExtractExecutableTask` we can directly inject `ArchiveOperations` removing the use of `getProject`

We have also added tests in `ConfigurationCacheTest` to prevent regressions and updated `GradleExecUtilsProjectSpec` to reflect the new usage of `GradleExecUtils`

==COMMIT_MSG==
Make `gradle-conjure` configuration cacheable for the `classes` task by removing the capturing of the `Project` type 
==COMMIT_MSG==


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureGeneratorTask.java
```diff
@@ -92,17 +102,17 @@ public void compileFiles() {
...
 
-            getProject().mkdir(thisOutputDirectory);
-
             List<String> generateCommand =
                     ImmutableList.of("generate", file.getAbsolutePath(), thisOutputDirectory.getAbsolutePath());
 
             GradleExecUtils.exec(
-                    getProject(),
+                    getExecOperations(),
+                    getBuildServiceRegistry(),
```
CRogers: I think a nicer to structure this would be to make a `GradleExecUtils` a managed type, where `ExecOperations` and `BuildServiceRegistry` are injected into it (like they currently are this task). This means we can just add it as `@Nested` and don't need to inject all it's dependencies (ie `ExecOperations` and `BuildServiceRegistry`) into each task.

FinlayRJW: updated to be managed and use `@Nested`

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,107 @@
...
+    def "classes task runs with configuration cache when applying conjure"() {
+        given:
+        buildFile << """
+            apply plugin: 'com.palantir.conjure'
+            apply plugin: 'java-library'
+
+            repositories {
+                mavenCentral()
+                mavenLocal()
+            }
```
CRogers: All this duplicated code should be extracted into the `setup` method.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,107 @@
...
+
+    def setup() {
+        definePluginOutsideOfPluginBlock = true
+        keepFiles = true
+    }
+
+    def "classes task runs with configuration cache when applying conjure"() {
+        given:
+        buildFile << """
+            apply plugin: 'com.palantir.conjure'
```
CRogers: I'm not sure these tests test the tasks are CC safe. I think we need to make the various conjure subprojects (eg `whatever-conjure-objects`) and add a conjure yml file for it to actually make the various generator/compilation tasks.

CRogers: It might be better in these cases to just convert the existing tests to IntegrationTestKitSpec like we've done elsewhere

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -316,15 +319,15 @@ class ConjurePluginTest extends IntegrationSpec {
...
+        BuildResult result = runTasks("compileConjure")
 
         then:
-        result.wasExecuted(':api:compileConjureObjects')
-        result.wasExecuted(':api:compileConjureJersey')
-        result.wasExecuted(':api:compileConjureTypeScript')
-        result.wasExecuted(':api:compileConjureUndertow')
-        result.wasExecuted(':api:compileConjureDialogue')
-        result.wasExecuted(':api:copyConjureSourcesIntoBuild')
+        result.task(':api:compileConjureObjects').outcome == TaskOutcome.UP_TO_DATE
```
FinlayRJW: not sure why these are all `UP_TO_DATE` rather than `SUCCESS` -> I'm guessing wasExecuted was checking if it was called not if the state was success - probably better to check the exact state now

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheTest.groovy
FinlayRJW: I've updated all the tests to use `IntegrationTestKitSpec` and where appropriate converted tests to use `runTasksWithConfigurationCache` this should (I think) cover everything hit by classes but not much else, need a FLUP to make all the rest of the tasks CC friendly.

Have confirmed these tests hit CC problems on develop

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureBasePluginIntegrationSpec.groovy
```diff
@@ -122,12 +129,12 @@ class ConjureBasePluginIntegrationSpec extends IntegrationSpec {
         when:
         file('src/main/conjure/api.yml') << API_YML
 
-        runTasksSuccessfully('compileIr')
+        runTasks('compileIr')
```
CRogers: Why is this one not `runTasksWithConfigurationCache`?

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
CRogers: Yeah, I think `IntegrationSpec`'s `wasExecuted` includes tasks that were "run" as part of the Gradle build, either up-to-date or successful/failed. This is better for tests.

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -316,15 +319,15 @@ class ConjurePluginTest extends IntegrationSpec {
...
+        result.task(':api:compileConjureJersey').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureTypeScript').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureUndertow').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureDialogue').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:copyConjureSourcesIntoBuild').outcome == TaskOutcome.SUCCESS
```
CRogers: Weird but I guess this was actually what was happening

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -17,16 +17,19 @@
...
+import nebula.test.IntegrationTestKitSpec
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.TaskOutcome
 import spock.lang.IgnoreIf
 import spock.lang.Unroll
 import spock.util.environment.RestoreSystemProperties
 
 @Unroll
-class ConjurePluginTest extends IntegrationSpec {
+class ConjurePluginTest extends IntegrationTestKitSpec {
```
CRogers: Is it expected none of these tests run with `--configuration-cache`?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
```diff
@@ -60,6 +48,28 @@ public void execute(Task task) {
     @Input
     public abstract Property<String> getPackageVersion();
 
+    @TaskAction
+    public final void cleanOutputDirectory() {
```
FinlayRJW: moved to task action - this passes all the tests but slightly worried this will happen to late now

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
FinlayRJW: I have updated every test that can to run with configuration cache this required fixes to `CompileConjureTypeScriptTask` and `GenerateConjureServiceDependenciesTask` 

A range of tests are still not CC compatible due to issues in `sls-packaging` and `BetterExec` -> I've marked this in the tests

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureBasePluginIntegrationSpec.groovy
FinlayRJW: this was due to the way I was running tests with configuration cache not clearing the cache after the two runs causing it to fail if you ran the same task twice - this is now resolved

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
```diff
@@ -60,6 +48,28 @@ public void execute(Task task) {
...
+        if (file.isDirectory()) {
+            File[] children = file.listFiles();
+            if (children != null) {
+                for (File child : children) {
+                    deleteRecursively(child);
+                }
+            }
+        }
+        file.delete();
+    }
```
CRogers: Import `commons-io` and use `FileUtils.deleteDirectory`.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
CRogers: I don't think having multiple `@TaskAction`s will keep a defined ordering. We should definitely continue to use `doFirst` here to ensure the clean runs first.

This is a potential errorprone check

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalGeneratorTask.java
```diff
@@ -34,43 +35,36 @@
...
-        return executablePath;
-    }
+    protected abstract RegularFileProperty getExecutablePath();
 
     @Input
-    public final MapProperty<String, Object> getOptions() {
-        return this.options;
+    protected abstract MapProperty<String, Object> getOptions();
+
+    // Set the path sensitivity of the sources, which would otherwise default to ABSOLUTE
```
CRogers, line 63: Crazy this is not already the default. I think ensuring every `@CacheableTask` has a `@PathSensitivity` on all `@InputFile`/`@InputFiles` is another error-prone we can add to the list.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
```diff
@@ -20,34 +20,22 @@
...
 import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.PathSensitivity;
+import org.gradle.api.tasks.TaskAction;
 
 @CacheableTask
 public abstract class CompileConjureTypeScriptTask extends ConjureGeneratorTask {
     public CompileConjureTypeScriptTask() {
-        Project project = getProject();
-        getPackageName().convention(getProject().provider(project::getName));
+        getPackageName().convention(getProject().provider(() -> getProject().getName()));
```
felixdesouza: I don't think we need to wrap the project name in a provider, you can't change it dynamically. The same is not true for the version however.

felixdesouza: We should also try to remove all references to `getProject` where there are actual alternatives i.e. inject `getProviderFactory` and use that instead.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalGeneratorTask.java
felixdesouza: It wouldn't really be safe for `RELATIVE` to be the default no?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java
```diff
@@ -144,8 +146,9 @@ public void invoke(Project project, String failedTo, List<String> unloggedArgs,
                     .addAll(loggedArgs)
                     .build();
 
-            ExecResult execResult = project.exec(execSpec -> {
-                project.getLogger().info("Running with args: {}", loggedArgs);
+            log.info("Running external process: {} with args: {}", executable.getName(), loggedArgs);
```
felixdesouza, line 151: ~let's bump this down to debug?~ nvm, it's the same as before

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalGeneratorTask.java
CRogers: When is ABSOLUTE ever the right choice if the task is also cacheable? It just means your task is never cacheable. Honestly, I don't see when RELATIVE would be dangerous except contrived situations. It's something Gradle has wished they had done by default but now can't: https://github.com/gradle/gradle/issues/10346

felixdesouza: agree, but it's more opt-in than not I guess, or maybe just putting the `@CacheableTask` is enough to indicate that?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractExecutableTask.java
```diff
@@ -48,12 +50,15 @@
 public abstract class ExtractExecutableTask extends Sync {
```
felixdesouza, line 50: FLUP, this should just be an artifact transform. I think we probably already have one, see `UnpackingArtifactTransformAction` from internal repository perhaps. You'd have to verify the output is the same or just adapt the consumer of this task

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExec.java
```diff
@@ -20,34 +20,37 @@
...
-                                @Override
-                                public void execute(BuildServiceSpec<Params> spec) {
-                                    spec.getParameters().getExecutable().set(executable);
-                                }
+                            (BuildServiceSpec<Params> spec) -> {
+                                spec.getParameters().getExecutable().set(executable);
                             })
                     .get()
-                    .invoke(project, failedTo, unloggedArgs, loggedArgs);
+                    .invoke(getExecOperations(), failedTo, unloggedArgs, loggedArgs);
```
felixdesouza: this is a bit weird? shouldn't you be able to inject `ExecOperations` into the build service?

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExec.java
```diff
@@ -20,34 +20,37 @@
...
                             // Executable name must be the cache key, neither the spec parameters
                             // nor the class are taken into account for caching.
                             "conjure-runner-" + executable,
                             ConjureRunnerResource.class,
-                            new Action<BuildServiceSpec<Params>>() {
-                                @Override
-                                public void execute(BuildServiceSpec<Params> spec) {
-                                    spec.getParameters().getExecutable().set(executable);
-                                }
+                            (BuildServiceSpec<Params> spec) -> {
```
felixdesouza: is there any reason to explicitly specify the types here?

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheSpec.groovy
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.conjure
+
+import nebula.test.IntegrationTestKitSpec
+import org.gradle.testkit.runner.BuildResult
+
+
+abstract class ConfigurationCacheSpec extends IntegrationTestKitSpec {
```
felixdesouza, line 23: nit: this isn't really a spec, because it doesn't contain any tests or typical setup/teardown methods. You want to extend `IntegrationTestKitBase` instead which has all the methods on it, and make this class a groovy trait so you're not having to force the hierarchy for every test class.

felixdesouza: You could probably leave it as is, and add a `setup` method, that calls into the parent `setup` method. But do some common things you do in all the other set up methods

```
definePluginOutsideOfPluginBlock = true
keepFiles = true
```

Then it becomes a spec

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -220,27 +223,27 @@ class ConjurePluginTest extends IntegrationSpec {
...
         when:
-        runTasksSuccessfully('compileJava')
+        runTasksWithConfigurationCache('compileJava')
 
         then:
-        fileExists(prefixPath(prefix, 'api-jersey/build/generated/sources/conjure-jersey/java/main'))
-        fileExists(prefixPath(prefix, 'api-objects/build/generated/sources/conjure-objects/java/main'))
-        fileExists(prefixPath(prefix, 'api-undertow/build/generated/sources/conjure-undertow/java/main'))
-        fileExists(prefixPath(prefix, 'api-dialogue/build/generated/sources/conjure-dialogue/java/main'))
+        new File(projectDir, prefixPath(prefix, 'api-jersey/build/generated/sources/conjure-jersey/java/main')).exists()
```
felixdesouza: let's just add the `fileExists` method to the `ConfigurationCacheSpec`

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -164,29 +167,29 @@ class ConjurePluginTest extends IntegrationSpec {
...
-        result2.wasUpToDate(prefixProject(prefix, 'api-objects:compileJava'))
-        result2.wasUpToDate(':api:compileConjureObjects')
-        result2.wasUpToDate(prefixProject(prefix, 'api-jersey:compileJava'))
-        result2.wasUpToDate(':api:compileConjureJersey')
-        result2.wasUpToDate(prefixProject(prefix, 'api-undertow:compileJava'))
-        result2.wasUpToDate(':api:compileConjureUndertow')
-        result2.wasUpToDate(prefixProject(prefix, 'api-dialogue:compileJava'))
-        result2.wasUpToDate(':api:compileConjureDialogue')
+        result.task(':extractConjureJava').outcome == TaskOutcome.SUCCESS
+        result.task(':' + prefixProject(prefix, 'api-objects:compileJava')).outcome == TaskOutcome.SUCCESS
```
felixdesouza: this is kinda grim from an api perspective, any new test will need to remember to add a `:`. Have `prefixProject` do the right thing and tack on a `:` at the beginning

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -164,29 +167,29 @@ class ConjurePluginTest extends IntegrationSpec {
...
+        result.task(':api:compileConjureObjects').outcome == TaskOutcome.SUCCESS
```
felixdesouza: I don't think these are the same thing. How nebula define it is basically did it appear in the list of tasks. Have a look at how it's defined in nebula:

https://github.com/nebula-plugins/nebula-test/blob/223f2002e52597011a3e0bcc950e1dc4aab69482/src/main/groovy/nebula/test/functional/internal/toolingapi/BuildLauncherBackedGradleHandle.groovy#L100-L105

felixdesouza: perhaps this is a reasonable stricter interpretation, just wanted to flag that they're not the same

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -250,39 +253,39 @@ class ConjurePluginTest extends IntegrationSpec {
...
-        result.wasUpToDate(':api:compileConjureDialogue')
-        result.wasUpToDate(':api:copyConjureSourcesIntoBuild')
-        result.wasUpToDate(':api:compileIr')
+        result.task(':api:compileConjureObjects').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureJersey').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureTypeScript').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureUndertow').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileConjureDialogue').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:copyConjureSourcesIntoBuild').outcome == TaskOutcome.UP_TO_DATE
+        result.task(':api:compileIr').outcome == TaskOutcome.UP_TO_DATE
```
felixdesouza: I think a better way to do this is:

```
result.tasks(TaskOutcome.UP_TO_DATE).containsAll(...)
```

there's just less visual noise

felixdesouza: the same applies to all the test classes

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
```diff
@@ -84,29 +86,30 @@ class ConjurePublishTypeScriptTest extends IntegrationSpec {
...
         directory('api/api-typescript/src/node_modules').exists()
 
         when:
-        ExecutionResult second = runTasksSuccessfully('-i', 'installTypeScriptDependencies')
+        BuildResult second = runTasksWithConfigurationCache('-i', 'installTypeScriptDependencies')
 
         then:
-        second.wasExecuted(':api:compileConjureTypeScript') // this should really be up-to-date, but something touches the output package.json which makes gradle re-run this
-        second.wasUpToDate(':api:installTypeScriptDependencies')
+        second.task(':api:compileConjureTypeScript').outcome == TaskOutcome.UP_TO_DATE // this should really be up-to-date, but something touches the output package.json which makes gradle re-run this
```
felixdesouza: hmm, something to consider is that your configuration cache runner, can mess with results in a way that's a bit subtle i.e. if you re-run twice - which your `runTasksWithConfigurationCache` is doing, and you were expecting a success, you might get an up-to-date.

In this case that I'm commenting on, the comment is saying it should be up-to-date, which it is now. But is it because the underlying issue is fixed, or because your configuration cache helper is skewing things?

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
```diff
@@ -152,34 +155,37 @@ class ConjurePublishTypeScriptTest extends IntegrationSpec {
...
+        result.task(':api:installTypeScriptDependencies').outcome == TaskOutcome.FAILED
 
         cleanup:
         server.shutdown()
     }
 
     def 'compiles TypeScript'() {
         when:
-        ExecutionResult result = runTasksSuccessfully(':api:compileTypeScript')
+        // Cannot run with configuration cache until BetterExec is fixed
```
felixdesouza: I think remove the comment, and put this as a FLUP so you fix it later

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/GradleExecUtilsProjectSpec.groovy
```diff
@@ -18,11 +18,32 @@ package com.palantir.gradle.conjure
 
 import nebula.test.ProjectSpec
 import org.assertj.core.api.Assertions
+import org.gradle.api.services.BuildServiceRegistry
+import org.gradle.process.ExecOperations
 
 class GradleExecUtilsProjectSpec extends ProjectSpec {
+
+    GradleExec createGradleExec() {
```
felixdesouza: same thing as before, you should be able to just instantiate this class instead of having to override it.

felixdesouza (commented):
I made a comment somewhere, but I think for things where you're checking whether something is up-to-date or successful, your utility method introduces a bit of ambiguity. It might be up-to-date because the cc runs things twice. Granted you return the first run, but you still run gradle twice.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalGeneratorTask.java
FinlayRJW: added to the errorprone list

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
FinlayRJW: I've moved this to where `CompileConjureTypeScriptTask` is registered so now we do:

```java
task.getPackageName().convention(project.getName());
task.getPackageVersion().convention(project.provider(() -> project.getVersion().toString()));
```

FinlayRJW: I think that feels tidier than doing it within the constructor

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ExtractExecutableTask.java
FinlayRJW: added as a FLUP -> https://pl.ntr/2v3

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExec.java
FinlayRJW: I originally did this as the second call of `invoke` is directly on `ConjureRunner` so thought it might be neater to just pass it into both - have changed to inject into the `ConjureRunnerResource` but pass into `ConjureRunner`

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExec.java
FinlayRJW: no - removed

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheSpec.groovy
FinlayRJW: cool have made into a proper spec and created a `fileExists` type - I wonder if it would be good to create our own general `ConfigurationCacheSpec` somewhere shared (utils?) that could have all this to save re-creating it each time?

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
FinlayRJW: so I had a think about this and you are right that this will cause issues - I've updated to now have two methods `runTasksWithConfigurationCacheAndCheck` which runs a task twice and checks this should only be used in tests where that task can be ran multiple times and `runTasksWithConfigurationCache` which just runs once with configuration cache and is not checking the output at all so can be used on all tests

FinlayRJW: so I think its only 100% safe to use `runTasksWithConfigurationCacheAndCheck` in cases where you are not checking the tasks state? - not sure if its okay in cases where you are checking a file / the output

FinlayRJW: I've only ran `runTasksWithConfigurationCacheAndCheck` in places where you only check the output as I'm pretty sure that is safe

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePublishTypeScriptTest.groovy
FinlayRJW: added as a FLUP

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/GradleExecUtilsProjectSpec.groovy
FinlayRJW: changed to use

```java
        ObjectFactory objectFactory = project.getObjects();
        GradleExec gradleExec = objectFactory.newInstance(GradleExec.class);
```

FinlayRJW:
> I made a comment somewhere, but I think for things where you're checking whether something is up-to-date or successful, your utility method introduces a bit of ambiguity. It might be up-to-date because the cc runs things twice. Granted you return the first run, but you still run gradle twice.

I've changed to only run once on most tests and only run twice in cases where we only check stdout (which im pretty sure is safe)

gradle-conjure/src/main/java/com/palantir/gradle/conjure/CompileConjureTypeScriptTask.java
FinlayRJW: converted to a `doFirst`

```java
    public CompileConjureTypeScriptTask() {
        doFirst(new Action<Task>() {
            @Override
            public void execute(Task _task) {
                File outputDir = getOutputDirectory().get().getAsFile();
                File nodeModulesDir = new File(outputDir, "node_modules");
                if (nodeModulesDir.exists()) {
                    try {
                        FileUtils.deleteDirectory(nodeModulesDir);
                    } catch (IOException e) {
                        throw new RuntimeException("Failed to delete node_modules directory", e);
                    }
                }
            }
        });
    }
```

FinlayRJW: whoops I was doing the complete opposite of what I should be doing! the old code deleted everything not under `node_modules` (`fileTree.exclude("node_modules/**/*");`) I was deleting `node_modules` - updated to:

```java
        doFirst(new Action<Task>() {
            @Override
            public void execute(Task _task) {
                File outputDir = getOutputDirectory().get().getAsFile();
                File[] files = outputDir.listFiles();
                if (files != null) {
                    Arrays.stream(files)
                            .filter(file -> !file.getName().equals("node_modules"))
                            .forEach(file -> {
                                try {
                                    if (file.isDirectory()) {
                                        FileUtils.deleteDirectory(file);
                                    } else {
                                        FileUtils.delete(file);
                                    }
                                } catch (IOException e) {
                                    throw new RuntimeException("Failed to delete: " + file, e);
                                }
                            });
                }
            }
        });
```

which should be CC safe and do what it used to

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
FinlayRJW: I think this is similar to here https://github.com/palantir/gradle-conjure/pull/1637#discussion_r2164273925 where we decided the new method is better as it is stricter

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
FinlayRJW: updated

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheSpec.groovy
```diff
@@ -0,0 +1,62 @@
...
```
kelvinou01, line 23: `ConfigurationCacheSpec` seems like a good idea! Should we put this in gradle-utils and use it across our other repositories?

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureBasePluginIntegrationSpec.groovy
```diff
@@ -79,7 +83,7 @@ class ConjureBasePluginIntegrationSpec extends IntegrationSpec {
         file('src/main/conjure/api.yml') << API_YML
 
         then:
-        runTasksSuccessfully('compileIr').standardOutput.contains '--verbose'
+        runTasksWithConfigurationCache('compileIr', '--info').output.contains('--verbose')
```
kelvinou01, line 86: nit: Let's do the execution itself in `when`, and checking in `then`? Also, let's extract the output into it's own variable, so the error message is cleaner if the test fails?

```groovy
when:
let output = runTasksWithConfigurationCache('compileIr', '--info').output

then:
output.contains('--verbose')
```

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureBasePluginIntegrationSpec.groovy
```diff
@@ -122,12 +124,12 @@ class ConjureBasePluginIntegrationSpec extends IntegrationSpec {
...
 
-        runTasksSuccessfully('compileIr')
+        runTasksWithConfigurationCache('compileIr')
         FileUtils.deleteDirectory(projectDir.toPath().resolve("build").toFile())
-        ExecutionResult result = runTasksSuccessfully('compileIr', '-i')
+        BuildResult result = runTasksWithConfigurationCache('compileIr', '-i')
 
         then:
-        result.standardOutput.contains "Task :compileIr FROM-CACHE"
+        result.output.contains "Task :compileIr FROM-CACHE"
```
kelvinou01: Same thing here, and for the other spock tests.

kelvinou01 (changes requested):

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheSpec.groovy
FinlayRJW: That was my idea - I've kept having todo the same thing repeatedly - not sure best place for it to live though

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureBasePluginIntegrationSpec.groovy
FinlayRJW: good idea - have done everywhere appropriate

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConfigurationCacheSpec.groovy
CRogers: Put it in gradle-plugin-testing

gradle-conjure/src/main/java/com/palantir/gradle/conjure/GradleExec.java
```diff
@@ -16,38 +16,37 @@
...
+    public abstract ExecOperations getExecOperations();
+
+    @Inject
+    public abstract BuildServiceRegistry getBuildServiceRegistry();
+
+    public final void exec(String failedTo, File executable, List<String> unloggedArgs, List<String> loggedArgs) {
 
-    static void exec(
-            Project project, String failedTo, File executable, List<String> unloggedArgs, List<String> loggedArgs) {
         if (gradleVersionHighEnough()) {
```
CRogers: Fwiw we can remove gradleVersionHighEnough here as everything is >=7.6.4

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureGeneratorTaskTest.groovy
```diff
@@ -86,20 +84,16 @@ class ConjureGeneratorTaskTest extends IntegrationSpec {
...
-        result2.wasExecuted(':api:compileConjure')
-        result2.wasExecuted(':api:compileConjureObjects')
-        result2.wasExecuted(':api:compileIr')
+        result2.tasks(TaskOutcome.SUCCESS)*.path.containsAll(':api:compileConjure', ':api:compileConjureObjects', ':api:compileIr')
 
         // java
-        !fileExists('api/api-objects/build/generated/sources/conjure-objects/java/main/test/test/api/StringExample.java')
-        fileExists('api/api-objects/build/generated/sources/conjure-objects/java/main/test/test/api/NewStringExample.java')
+        !fileExists( 'api/api-objects/build/generated/sources/conjure-objects/java/main/test/test/api/StringExample.java')
+        fileExists( 'api/api-objects/build/generated/sources/conjure-objects/java/main/test/test/api/NewStringExample.java')
```
CRogers: I'd run a find replace on all the files from `fileExists( '` -> `fileExists('`

CRogers:
:+1:

PR #1648, opened 2025-06-25
Add validation-api dependency to jersey projects when using NotNull

aldexis:
## Before this PR
When using [`requireNotNullAuthAndBodyParams`](https://github.com/palantir/conjure-java/blob/bd3dce573b5d92f6efd29f30a7c013f779030c91/conjure-java-core/src/main/java/com/palantir/conjure/java/Options.java#L37-L44) and generating the jersey subprojects, this would use `javax.validation.constraints.NotNull` (if using javax) or `jakarta.validation.constraints.NotNull` (if using jakarta)

However, we would not be adding the dependency to the jersey subproject, which would cause a compilation failure, and require users to manually add the dependency to their project.

## After this PR
==COMMIT_MSG==
Add validation-api dependency to jersey projects when using NotNull
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -117,6 +117,14 @@ public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
         if (Dependencies.isJersey(extension.getJava())) {
             project.getDependencies()
                     .add("api", useJakarta ? Dependencies.JAXRS_API_JAKARTA : Dependencies.JAXRS_API_JAVAX);
+            if (Dependencies.isNotNullAuthAndBodyParams(extension.getJava())) {
+                project.getDependencies()
+                        .add(
+                                "implementation",
+                                useJakarta
+                                        ? Dependencies.JAXRS_VALIDATION_API_JAKARTA
+                                        : Dependencies.JAXRS_VALIDATION_API_JAVAX);
+            }
         }
         if (Dependencies.isDialogue(extension.getJava())) {
             project.getDependencies().add("api", Dependencies.DIALOGUE_TARGET);
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
@@ -25,6 +25,8 @@ final class Dependencies {
     static final String ANNOTATION_API_JAVAX = "javax.annotation:javax.annotation-api:1.3.2";
     static final String JAXRS_API_JAKARTA = "jakarta.ws.rs:jakarta.ws.rs-api:3.0.0";
     static final String JAXRS_API_JAVAX = "javax.ws.rs:javax.ws.rs-api:2.1.1";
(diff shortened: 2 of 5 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -117,6 +117,14 @@ private static void configureDependencies(Project project, ConjureExtension exte
...
             project.getDependencies()
                     .add("api", useJakarta ? Dependencies.JAXRS_API_JAKARTA : Dependencies.JAXRS_API_JAVAX);
+            if (Dependencies.isNotNullAuthAndBodyParams(extension.getJava())) {
+                project.getDependencies()
+                        .add(
+                                "implementation",
+                                useJakarta
+                                        ? Dependencies.JAXRS_VALIDATION_API_JAKARTA
+                                        : Dependencies.JAXRS_VALIDATION_API_JAVAX);
+            }
```
aldexis, line 127: I don't like this duplicated piece of code, but I'm merely following the existing code. Refactoring it to share the dependencies' setup logic is outside of the scope of this PR.

Similarly, I wanted to add similar tests for the conjure-java-local setup, but it turns out that we don't have a good way to test anything but object generation right, as this is generating from https://github.com/palantir/conjure/blob/master/conjure-api/src/main/conjure/conjure-api.yml, which doesn't have a service to generate. Adding tests for this is also outside of the scope of this PR.

gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
```diff
@@ -36,24 +38,37 @@ final class Dependencies {
...
+        return isOption(options, DIALOGUE);
     }
 
     static boolean isUndertow(GeneratorOptions options) {
-        return options.has(UNDERTOW) && Boolean.TRUE.equals(options.get(UNDERTOW));
+        return isOption(options, UNDERTOW);
+    }
+
+    private static boolean isOption(GeneratorOptions options, String optionName) {
+        return options.has(optionName) && Boolean.TRUE.equals(options.get(optionName));
```
aldexis, line 71: Small drive-by refactor to share the logic of extracting a boolean option

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjurePluginTest.groovy
```diff
@@ -159,6 +159,71 @@ class ConjurePluginTest extends IntegrationSpec {
...
+
+        then:
+        result.wasExecuted(prefixProject(prefix, 'api-objects:compileJava'))
+        result.wasExecuted(':api:compileConjureObjects')
+        result.wasExecuted(prefixProject(prefix, 'api-jersey:compileJava'))
+        result.wasExecuted(':api:compileConjureJersey')
+        result.wasExecuted(prefixProject(prefix, 'api-undertow:compileJava'))
+        result.wasExecuted(':api:compileConjureUndertow')
+        result.wasExecuted(prefixProject(prefix, 'api-dialogue:compileJava'))
+        result.wasExecuted(':api:compileConjureDialogue')
```
aldexis, line 186: There is likely some refactor that can be done to share some of the test setup, but this should be done separately from this PR as well

bjlaub (approved):
thanks!

PR #1649, opened 2025-06-25
Share dependency setup between normal and local conjure generation plugins

aldexis:
## Before this PR
Follow-up on https://github.com/palantir/gradle-conjure/pull/1648

I realized that we had somewhat duplicated logic for setting up the dependencies on a conjure subproject, between the normal plugin and the conjure-java-local one.

It turns out that at least parts could be shared

## After this PR
==COMMIT_MSG==
Share dependency setup between normal and local conjure generation plugins
==COMMIT_MSG==

### DIFF
--- a/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
+++ b/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
@@ -115,22 +115,13 @@ public final class ConjureJavaLocalCodegenPlugin implements Plugin<Project> {
                         "compileOnly",
                         useJakarta ? Dependencies.ANNOTATION_API_JAKARTA : Dependencies.ANNOTATION_API_JAVAX);
         if (Dependencies.isJersey(extension.getJava())) {
-            project.getDependencies()
-                    .add("api", useJakarta ? Dependencies.JAXRS_API_JAKARTA : Dependencies.JAXRS_API_JAVAX);
-            if (Dependencies.isNotNullAuthAndBodyParams(extension.getJava())) {
-                project.getDependencies()
-                        .add(
-                                "implementation",
-                                useJakarta
-                                        ? Dependencies.JAXRS_VALIDATION_API_JAKARTA
-                                        : Dependencies.JAXRS_VALIDATION_API_JAVAX);
-            }
+            ConjurePlugin.setupJerseyProject(project, extension::getJava);
         }
         if (Dependencies.isDialogue(extension.getJava())) {
-            project.getDependencies().add("api", Dependencies.DIALOGUE_TARGET);
+            ConjurePlugin.setupDialogueProject(project, extension::getJava);
         }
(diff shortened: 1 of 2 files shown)


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -115,22 +115,13 @@ private static void configureDependencies(Project project, ConjureExtension exte
                         "compileOnly",
                         useJakarta ? Dependencies.ANNOTATION_API_JAKARTA : Dependencies.ANNOTATION_API_JAVAX);
```
aldexis, line 116: I'm not entirely sure why the above dependencies are always applied. I tried to figure out whether they could be useful specifically for conjure-java-local (considering they are not set up in the same way for normal conjure projects), but I couldn't manage to figure it out.

In normal conjure plugin,
```
        project.getDependencies()
                .add(
                        "compileOnly",
                        useJakarta ? Dependencies.ANNOTATION_API_JAKARTA : Dependencies.ANNOTATION_API_JAVAX);
```
is [only done for the jersey projects](https://github.com/palantir/gradle-conjure/blob/4c21b4d91ee7d6d39884f531ffd5688d39e47bf6/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java#L299-L302) and 
```
        project.getDependencies().add("api", Dependencies.CONJURE_JAVA_LIB);
        project.getDependencies().add("api", Dependencies.JETBRAINS_ANNOTATIONS);
```
is [only done for the objects project](https://github.com/palantir/gradle-conjure/blob/4c21b4d91ee7d6d39884f531ffd5688d39e47bf6/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java#L279-L280)

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -280,11 +280,11 @@ private static void setupObjectsProject(Project project, Supplier<GeneratorOptio
         project.getDependencies().add("api", Dependencies.JETBRAINS_ANNOTATIONS);
     }
 
-    private static void setupDialogueProject(Project project, Supplier<GeneratorOptions> _optionsSupplier) {
+    static void setupDialogueProject(Project project, Supplier<GeneratorOptions> _optionsSupplier) {
```
aldexis, line 283: Followed the pattern of `ignoreFromCheckUnusedDependencies` from below (which is also made package-private and [reused in ConjureJavaLocalCodegenPlugin](https://github.com/palantir/gradle-conjure/blob/4c21b4d91ee7d6d39884f531ffd5688d39e47bf6/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java#L142), same as a couple of other methods), though I believe this all should probably be extracted to a common class

aldexis:
This is superseded by https://github.com/palantir/gradle-conjure/pull/1652

PR #1652, opened 2025-06-26
Disable checkImplicitDependencies by default on conjure subprojects and make existing dependencies explicit

aldexis:
## Before this PR
In the current state of things, the gradle-conjure plugin is not properly setting up dependencies for the conjure subprojects.
Specifically, it does not pass the [`checkImplicitDependencies`](https://github.com/palantir/gradle-baseline/blob/1d5024af928c7c3353d03d9e4a742fd1fe5451f5/gradle-baseline-java/src/main/java/com/palantir/baseline/tasks/CheckImplicitDependenciesTask.java) task defined by [BaselineExactDependencies](https://github.com/palantir/gradle-baseline/blob/1d5024af928c7c3353d03d9e4a742fd1fe5451f5/gradle-baseline-java/src/main/java/com/palantir/baseline/plugins/BaselineExactDependencies.java).

This means that if users want to wire up e.g. `tasks.check.dependsOn checkImplicitDependencies`, they also have to manually disable it within conjure subprojects.

We _should_ define the dependencies explicitly and have this task pass. However, due to the way this plugin works, we _cannot_ guarantee this will work.

This is due to the fact that the conjure-java generator may upgrade ahead of the gradle-conjure plugin itself, leading to the following case:
- Say `gradle-conjure` currently defines the exact dependencies appropriately and a consumer repository wired up the `checkImplicitDependencies` task to run on `check`
- `conjure-java` code generation adds a dependency on a new type, which happens to come from an existing transitive dependency
  - This in itself shouldn't be blocking upgrades, considering the dependency is still on the runtime classpath.
- `gradle-conjure` plugin isn't updated yet, but `conjure-java` gets bumped in a consumer repository
- The `checkImplicitDependencies` task then starts failing, because of the new transitive dependency.
- This blocks `conjure-java` upgrades until `gradle-conjure` is fixed and upgraded as well, or until someone comes and manually disables the check task
  - Note: this may even be worse, because `checkUnusedDependencies` may require you to upgrade both in lockstep, since the new dependency is unused in previous generator versions

Instead, it's better to simply disable the `checkImplicitDependencies` task on conjure subprojects.

To ensure we _do_ still have explicit dependencies as best as we can, we should also have tests that verify the dependencies are explicit.

## After this PR

==COMMIT_MSG==
Disable checkImplicitDependencies by default on conjure subprojects and make existing dependencies explicit
==COMMIT_MSG==


gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureJavaLocalCodegenPlugin.java
```diff
@@ -107,30 +107,17 @@ private static void setupSubprojects(
     }
 
     private static void configureDependencies(Project project, ConjureExtension extension) {
-        project.getDependencies().add("api", Dependencies.CONJURE_JAVA_LIB);
```
bjlaub, line 112: why do these now only get applied if `Dependencies.isObjects` is true?

aldexis: I believe they aren't actually used by the others, and decided it was better to be consistent with non-local codegen. I'll double-check with a local publication of the plugin as well (I think I already had, but will reverify since it's been a while)

aldexis: @bjlaub I've tested locally on a couple of internal repos (including using normal conjure as well as conjure-java-local), which showed that writing the locks and compiling everything was fine. I'd like to test with an RC as well, if everything else looks good to you

bjlaub (dismissed):

gradle-conjure/src/main/java/com/palantir/gradle/conjure/Dependencies.java
```diff
@@ -31,14 +34,36 @@ final class Dependencies {
...
     static final String DIALOGUE_TARGET = "com.palantir.dialogue:dialogue-target:3.135.0";
+
+    static final String GUAVA = "com.google.guava:guava:33.4.8-jre";
+    static final String UNDERTOW_CORE = "io.undertow:undertow-core:2.3.18.Final";
+    static final String JACKSON_ANNOTATIONS = "com.fasterxml.jackson.core:jackson-annotations:2.18.3";
+    static final String JACKSON_DATABIND = "com.fasterxml.jackson.core:jackson-databind:2.18.3";
+    static final String FINDBUGS = "com.google.code.findbugs:jsr305:3.0.2";
+    static final String ERROR_PRONE_ANNOTATIONS = "com.google.errorprone:error_prone_annotations:2.38.0";
+    static final String SAFE_LOGGING_PRECONDITIONS = "com.palantir.safe-logging:preconditions:3.9.0";
+    static final String SAFE_LOGGING = "com.palantir.safe-logging:safe-logging:3.9.0";
```
aldexis, line 45: To check: we may also need a dependency on conjure-java-runtime-api somewhere to depend on RemoteException? (for dialogue-generated clients)

gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java
```diff
@@ -329,6 +300,16 @@ static void ignoreFromCheckUnusedDependencies(Project proj) {
...
+
+            // We disable the wrapper checkImplicitDependencies task in conjure subprojects, so users can configure the
+            //   global check task to depend on it, even if the conjure generator version they use happens to
+            //   implicitly depend on transitive libraries that the current gradle plugin version does not declare
+            //   explicitly.
+            // We do still try to set up dependencies explicitly, and cover it in ConjurePluginTest, but it could
+            //   otherwise create problems when versions of the generator and plugin are not in sync.
+            proj.getTasks().withType(checkImplicitDependenciesParentTask).configureEach(task -> {
+                task.setEnabled(false);
+            });
```
aldexis, line 312: This seems to not work exactly the way I intended it too, as I see e.g. the `checkImplicitDependenciesMain` tasks still be run. I'll have to dig deeper into this, but likely won't have time for now. I'll get back to this once I'm back from holidays

PR #1657, opened 2025-07-03
Introduce `getProductDependenciesTransformers` to `ConjureProductDependenciesExtension` to help unblock configuration cache

FinlayRJW:
## Before this PR
Previously there was no mechanism for other plugins to safely modify the `ConjureProductDependenciesExtension.getProductDependencies` this resulted in other plugins just modifying the state of `getProductDependencies` on the extension directly. This used to work from dumb luck but now we are rolling configuration cache out the `convertDependencies` method is called before other plugins have a change to modify the state

## After this PR
Fundamentally this is meant to be a light touch PR we want to unblock configuration cache without having to do a massive re-write of `gradle-conjure` at some point we will be force to comeback and make a lot of changes but for now I'm trying to keep it minimal

If we where to build this from the ground up we should have used the `NamedDomainContainer` but this forces you to change the was you set `serviceDependencies` in your `build.gradle` so instead we settled on adding a `getProductDependenciesTransformers` method on the extension that allows other plugins to add methods that modify the `ServiceDependency` as late as possible

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?
I couldn't figure out how to get a test working in the `IntegrationSpec` so I've added a small java test for the extension this can be a FLUP if it is critical

This does introduce a few breaks! 
- `getProductDependencies` and `getEndpointVersions` are now `SetProperties` 
- `ConjureProductDependenciesExtension` is abstract 

This should not be much of a problem we only seriously use this extension in one internal plugin -> https://pl.ntr/2v9 a PR to fix this there will follow soon


gradle-conjure-api/src/main/java/com/palantir/gradle/conjure/api/ConjureProductDependenciesExtension.java
```diff
@@ -16,53 +16,37 @@
...
     }
 
-    public final void serviceDependency(@DelegatesTo(ServiceDependency.class) Closure<ServiceDependency> closure) {
-        ServiceDependency serviceDependency = new ServiceDependency();
-        closure.setDelegate(serviceDependency);
-        closure.call();
-        productDependencies.add(serviceDependency);
-    }
+    public abstract Property<Function<Set<ServiceDependency>, Set<ServiceDependency>>>
+            getProductDependenciesTransformer();
```
CRogers: I think we should allow multiple transformers on here. So it would be something like a `ListProperty<...>` that people `.add` to.

CRogers: Otherwise if we end up with two plugins trying to set this they will fight eachother.

CRogers: I actually wonder if we should hide all this and use a `NamedDomainObjectSet<ServiceDependencies> getProductDependencies()`... This would enable other plugins to call `getProductDependencies().configureEach` which is basically implementing what we have here. But it would require the `ServiceDependency`s not being realized until we need them to (currently they need to get realised right away so people can set the productGroup and productName), or changing the fields on `ServiceDependency` to be `Property`s.

.palantir/revapi.yml
```diff
@@ -49,3 +49,149 @@ acceptedBreaks:
       old: "method java.util.Optional<java.lang.String> com.palantir.gradle.conjure.api.EndpointVersionBound::getMaxVersion()"
       new: "method java.lang.String com.palantir.gradle.conjure.api.EndpointVersionBound::getMaxVersion()"
       justification: "Reverting a change."
+  "5.57.0":
+    com.palantir.gradle.conjure:gradle-conjure-api:
+    - code: "java.annotation.removed"
+      old: "method void com.palantir.gradle.conjure.api.ConjureProductDependenciesExtension::<init>(org.gradle.api.Project)"
+      new: "method void com.palantir.gradle.conjure.api.ConjureProductDependenciesExtension::<init>()"
+      justification: "ConjureProductDependenciesExtension has been modernised to use\
```
FinlayRJW, line 57: Is this a good enough justification? should I do a separate one for each break?

gradle-conjure/src/test/java/com/palantir/gradle/conjure/ConjureProductDependenciesExtensionTest.java
```diff
@@ -0,0 +1,134 @@
...
+import org.gradle.testfixtures.ProjectBuilder;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ConjureProductDependenciesExtensionTest {
+
+    private ConjureProductDependenciesExtension extension;
+
+    @BeforeEach
+    void beforeEach() {
```
FinlayRJW: I really tried to write a test for this in `ConjureServiceDependencyTest` I added:

```groovy
    def "custom productDependenciesTransformer filters dependencies (integration)"() {
        //language=gradle
        file('api/build.gradle') << '''
        import java.util.function.Function
            
        serviceDependencies {
            productDependenciesTransformers.add(new Function<Set, Set>() {
                @Override
                Set apply(Set deps) {
                    println "--> EXECUTING MY TRANSFORMER on ${deps.size()} dependencies!"
                    def modified = deps.collect { dep ->
                        dep.minimumVersion = "1.1.1"
                        dep
                    }
                    return modified as Set
                }
            })
        
            serviceDependency {
                productGroup = "bar"
                productName = "foo"
                recommendedVersion = "1.0.0"
                maximumVersion = "2.0.0"
            }
        }
        '''.stripIndent(true)
        when:
        runTasksSuccessfully('compileConjure')

        then:
        def deps = file('api/build/service-dependencies.json').text
        deps.contains('"minimumVersion":"1.1.1"')
    }
```

But this failed todo anything even though this worked when I tested it on a real service??

FinlayRJW: try in jar

PR #1681, opened 2025-07-28
Use `com.palantir.gradle.plugintesting.ConfigurationCacheSpec` instead of our own implementation 

FinlayRJW:
## Before this PR
Had our own implementation of `com.palantir.gradle.plugintesting.ConfigurationCacheSpec` 

## After this PR
Now we use the same shared spec rather than having it re-implemented everywhere

==COMMIT_MSG==
Use `com.palantir.gradle.plugintesting.ConfigurationCacheSpec` instead of our own implementation 
==COMMIT_MSG==


gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/FileExists.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.conjure
+
+trait FileExists {
+    Boolean fileExists(String path) {
+        return new File(projectDir, path).exists()
+    }
+}
```
kelvinou01: Nice! I love this. 

nit: let's add a whitespace at EOF?

kelvinou01 (dismissed):
lets go!

kelvinou01 (approved):

PR #1954, opened 2026-04-07
Move conjure TypeScript output from `src/` to `build/generated/sources/`

FinlayRJW:
## Before this PR

The conjure TypeScript generator outputs to the typescript subproject's `src/` directory ([originally introduced](https://github.com/palantir/gradle-conjure/pull/792)). When consumer repos broadly apply `java-library` to all subprojects, the typescript subproject inherits Java tasks whose input/output directories overlap with conjure's `src/` output. On Gradle 9, this causes implicit dependency validation errors for `processResources`, `compileJava`, etc.

#1949 partially addressed this by disabling `JavaCompile` tasks, but other task types like `processResources` still trigger the same error — disabling them one-by-one is whack-a-mole.

## After this PR

==CHANGELOG_MSG==
The conjure TypeScript generator now outputs to `build/generated/sources/conjure-typescript/` instead of `src/`.
==CHANGELOG_MSG==

Moves the output to `build/generated/sources/conjure-typescript/`, matching the Java generators' convention. No directory overlap with Java source sets means no implicit dependency errors, regardless of what plugins are applied. 

Also removes the `JavaCompile` disable workaround from #1949 and the `.gitignore` generation (unnecessary since `build/` is already gitignored).

## Possible downsides?

Breaking change for repos assume the location of the typescript output - although I think this was always meant as more of an implementation detail than a location for users to interact with

## Testing

All tests passing, fixes Gradle 9 issues

### DIFF
--- a/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureLocalPluginTest.groovy
+++ b/gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureLocalPluginTest.groovy
@@ -104,7 +104,7 @@ class ConjureLocalPluginTest extends ConfigurationCacheSpec implements FileExist
         then:
         result.tasks(TaskOutcome.SUCCESS)*.path.containsAll(":generateTypeScript", ":generatePython")
 
-        fileExists('typescript/src/conjure-api/index.ts')
+        fileExists('typescript/build/generated/sources/conjure-typescript/conjure-api/index.ts')
         fileExists('python/python/conjure-api/conjure_spec/__init__.py')
     }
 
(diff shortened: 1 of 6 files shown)


FinlayRJW:
works for gradle 9 - https://pl.ntr/2Ax

gradle-conjure/src/test/groovy/com/palantir/gradle/conjure/ConjureLocalPluginTest.groovy
```diff
@@ -104,7 +104,7 @@ class ConjureLocalPluginTest extends ConfigurationCacheSpec implements FileExist
         then:
         result.tasks(TaskOutcome.SUCCESS)*.path.containsAll(":generateTypeScript", ":generatePython")
 
-        fileExists('typescript/src/conjure-api/index.ts')
+        fileExists('typescript/build/generated/sources/conjure-typescript/conjure-api/index.ts')
```
crogoz, line 107: would the old generated code in `src/` be deleted and written instead in `build/generated/sources` ?

FinlayRJW: I don't think it would be deleted as it was never checked-in, but it would not be written too `build/generated/sources`

crogoz (approved):

